PR #28, opened 2016-02-02
Missing memory in old computer resume, added total and used

mckaygerhard:
in the hardinfo computer module resumen the memory amount are missing, so added here!

### DIFF
--- a/modules/computer.c
+++ b/modules/computer.c
@@ -364,6 +364,7 @@ gchar *callback_summary()
     gchar *input_devices, *printers;
     gchar *storage_devices, *summary;
     gchar *virt;
+    MemoryInfo *mi = computer_get_memory();
     
     processor_name  = module_call_method("devices::getProcessorName");
     alsa_cards      = computer_get_alsacards(computer);
@@ -378,7 +379,7 @@ gchar *callback_summary()
 			      "#ReloadInterval=5000\n"
 			      "[Computer]\n"
 			      "Processor=%s\n"
-			      "Memory=...\n"
+			      "Memory=%s\n"
 			      "Machine Type=%s\n"
 			      "Operating System=%s\n"
 			      "User Name=%s\n"
@@ -392,6 +393,7 @@ gchar *callback_summary()
 			      "\n%s\n"
 			      "\n%s\n"),
 			      processor_name,
+			      g_strdup_printf(_("%dMB (%dMB used)"), mi->total, mi->used),
 			      virt,
 			      computer->os->distro,
 			      computer->os->username,
@@ -408,6 +410,7 @@ gchar *callback_summary()
     g_free(printers);
     g_free(storage_devices);
     g_free(virt);
+    g_free(mi);
 
     return summary;
 }


lpereira:
Memory information is updated very second already; `UpdateInterval$Memory=1000` in the `$ShellParam$` assures that.

mckaygerhard:
well but if u see that tab of information the screen show "memory " and not show amount .. that's why i pull this its a little change only

lpereira:
If it's not updating, it's a bug somewhere else, likely in the shell rather than the computer module. I find it very unlikely, as that is some well-tested part of the code and it works on every machine I've tried HardInfo on.

Are you using HardInfo with some translation? If so, it's possible that it's been poorly translated and the `UpdateInterval$Memory=1000` has not been translated to match the `Memory` key.

mckaygerhard:
umm i'm using hardinfo with es translation, are u sure?, i'll test then tomorrow with no translation to see what u refers and then close or feedback here ..

mckaygerhard:
well u have right in part, using different lang this king of memory info th show are broken

using this for launch:
```
export LANG=C;export LANGUAGE=C;export GDM_LANG=C;export LC_ALL=C;hardinfo
```
hardinfo correctly show so i'll open a issue and let this pull request opened until u decided what are best to do

mckaygerhard:
i confirmed with the issue comment #38#issuecomment-269538167 ( https://github.com/lpereira/hardinfo/issues/38#issuecomment-269538167) i quoted:

> Look at the Memory, it can't be translated, because the value disappears with it. There is the comment "//Load Average is not updated if locale is not C, switch locale to C" in /modules/computer.c (https://github.com/lpereira/hardinfo/blob/master/modules/computer.c). 

this patch its not elegant but for now its works and make it tranlatable, other way its skip this line in po files

lpereira:
With this patch, the memory information will be obtained only once. It's not the right fix.

mckaygerhard:
umm, so unstranlated maybe the best "for now" ? or made two parts and then join will work?

lpereira:
Here's an example of how to translate an automatically-updating field. Original, in English:

    [$ShellParam$]
    UpdateInterval$SomeField=1000
    [Some Group]
    SomeField=...

Translated:

    [$ShellParam$]
    UpdateInterval$Watermelon=1000
    [Pineapple]
    Watermelon=...

Then, edit the `hi_get_field()` for that module, and instead of calling `g_str_equal(field, "SomeField")`, call `g_str_equal(field, _("SomeField"))`. "SomeField" will be available as a translatable string, and if you translate that to "Watermelon", it should show up in the UI translated.

mckaygerhard:
thanks for the tip, i'll test in next days and post feedback here

TotalCaesar659:
After last commit (https://github.com/lpereira/hardinfo/commit/79a2c4d8d1589fa87e967f752686a090419b70ef), first label "Memory" becomes translatable, but not another ones.
2016-12-29-235500_1280x974_scrot
2016-12-29-235524_1280x974_scrot

lpereira:
There might be a bug in the shell. I'll try to look at that code later. (HardInfo is one of the first programs I wrote in C, so the code is far from being maintainable...)

lpereira:
I'm closing this bug report as the original bug has been fixed -- let's keep the conversation about the untranslatable strings in #38.

PR #29, opened 2016-02-02
fix null vendor in display and null version, the strings in get_srt ...

mckaygerhard:
fix null vendor in display and null version, 
strings in get_srt must be translated AFTER getting and putting in pointers


lpereira:
Merged, thank you!

PR #34, opened 2016-06-11
German translation // TiB and PiB for larger drives

theHacker:


lpereira:
Merged, vielen Danke!

PR #42, opened 2017-01-10
Added saving window size and maximized state to config file after closing

TotalCaesar659:


lpereira:
I totally missed this patch.  I'm sorry, but I don't think this is a good idea; this is the sort of thing that the window manager should do.  Saving and restoring the window position is tricky, specially when you're used to plugging a laptop into monitors of different resolutions.

TotalCaesar659:
OK, you're the boss, but do you have a plan for making in as option in preferences, if not automatically, as I did? It would be very comfortable not to change resolution every time. I had to make this patch because of it.

mckaygerhard:
i think that @lpereira said its that its not good an option that saves the window properties, as said, i have some problems when conected adtional monitors, and change sesion managers, then when open the apps that "saves" some windows properties, got a crash or invalid sizes .. etc.. this kind of task must be done by a more complex windo manager, openbox behavior its not same as mate behavior...

TotalCaesar659:
Oh, thanks for the explanation. But if author changes his mind, he's welcome for using this patch.

bp0:
Is old, conflicts, and may not be a good idea. Closing.

PR #47, opened 2017-04-09
Power Management flags list

bp0:
Like the Bugs list, there is a now a list for Power Management features. The first patch adds handling for that feature list.
The second patch is just an additional code comment for a "bug."
The third patch simply reorders the feature lists so that the very long Capabilities list is last.

### DIFF
--- a/includes/x86/processor-platform.h
+++ b/includes/x86/processor-platform.h
@@ -35,6 +35,7 @@ struct _Processor {
     gchar *vendor_id;
     gchar *flags;
     gchar *bugs;
+    gchar *pm;             /* power management features */
     gint cache_size;
     gfloat bogomips, cpu_mhz;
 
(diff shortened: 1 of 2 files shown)


mckaygerhard:
umm i not tested very well these "new" features, this programs are not only used on "most up to fashion" kernels and linux, also its still using in olders and low end hardware.. 

i mean, many the kworl series uses older kernel modules only available and working with v4l 1 api or some hardware such VIA motherboards does not have OpenGL capabilities in currently Xorg versions due only have DRI v1 code.. all of those still using older linux'es due their "capabilities" and if hardinfo only will work in "modern and up to date fashion" way .. i dont think so its a good idea.. for that the i used a up to date dekstop.. with up to date well-knowed program

bp0:
With the new commit, there is support for making a bugs list from the old set of bug flags, and handling the case where the Power Management flag list is not given. This should handle any old kernels.

lpereira:
Merged, thanks!

PR #52, opened 2017-04-10
More "feature" flags to bug workarounds

bp0:
Add some more bug "feature" flags from old kernels to the bug
workaround list generated if no such list was available in
/proc/cpuinfo.

Also, fixes a crash if the first bug is not fdiv.


lpereira:
Merged, thanks!

PR #61, opened 2017-05-22
Get board information for Raspberry Pis

bp0:
get_motherboard() will now return something for Raspberry Pi boards.
example:
"Raspberry Pi 3 (Q1 2016) pcb-rev:1.2 soc:BCM2835 mem:1GB mfg-by:Sony"

It can be seen in the GUI by selecting "Computer".

### DIFF
--- a/modules/devices.c
+++ b/modules/devices.c
@@ -222,7 +222,7 @@ static struct {
   { "a01040",    "Unknown",    "2 Model B",           "1.0",    "1GB",      "Sony" },
   { "a01041",    "Q1 2015",    "2 Model B",           "1.1",    "1GB",      "Sony" },
   { "a21041",    "Q1 2015",    "2 Model B",           "1.1",    "1GB",      "Embest" },
-  { "a22042",    "Q3 2016",    "2 Model B",           "1.2",    "1GB",    "Embest" },  /* (with BCM2837) */
+  { "a22042",    "Q3 2016",    "2 Model B",           "1.2",    "1GB",      "Embest" },  /* (with BCM2837) */
   { "900021",    "Q3 2016",    "A+",                  "1.1",    "512MB",    "Sony" },
   { "900032",    "Q2 2016?",    "B+",                 "1.2",    "512MB",    "Sony" },
   { "900092",    "Q4 2015",    "Zero",                "1.2",    "512MB",    "Sony" },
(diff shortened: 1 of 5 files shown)


bp0:
Should note that current linux reports BCM2835 as SOC for all Raspberry Pis, even if it is BCM2836 or BCM2837.

mckaygerhard:
the patch its not good, has some problems:
* assume only two archs: x86 and arm (threated as raspberry due assumed)
* has hardcoded features, and hardware may change in time by user modding

in my opinion, it's not good of merge

bp0:
@mckaygerhard 

> assume only two archs: x86 and arm (threated as raspberry due assumed)

Previously assumed a single arch: x86. This would be an expansion of capability.

> has hardcoded features, and hardware may change in time by user modding

I don't completely understand what you mean here. The board is the board even if the pi is modded. The hard-coded table can be expanded with new models, or "Unknown" will be returned. Previously "Unknown" was returned for all of them. There is also a new patch that will use the model from device tree for non-x86. If it starts with "Raspberry Pi", then the rpi revision is decoded. This also will give some information of future rpis, even before they are added to the detailed table.

mckaygerhard:
i only see that the  scan_dmi(FALSE) was hardcode to x86

the models list are hardcode for raspberry, its better get only "arm based" and the rest from the system.. raspberry its only a implementation of a arm based hardware

modules/devices.c
```diff
@@ -195,12 +195,142 @@ gchar *get_memory_total(void)
...
+  { "920093",    "Q4 2016?",   "Zero",                "1.3",    "512MB",    "Embest" },
+  { "9000c1",    "Q1 2017",    "Zero W",              "1.1",    "512MB",    "Sony" },
+  { "a02082",    "Q1 2016",    "3 Model B",           "1.2",    "1GB",      "Sony" },
+  { "a020a0",    "Q1 2017",    "Compute Module 3 or CM3 Lite",    "1.0",    "1GB",    "Sony" },
+  { "a22082",    "Q1 2016",    "3 Model B",           "1.2",    "1GB",    "Embest" },
+  { "a32082",    "Q4 2016",    "3 Model B",           "1.2",    "1GB",    "Sony Japan" },
+  { NULL, NULL, NULL, NULL, NULL, NULL }
+};
+
+gchar *dt_get_model() {
```
lpereira: You can use `g_file_get_contents()` instead of everything inside this function.  Please also make this a `static` function so it's not exported.

modules/devices.c
```diff
@@ -195,12 +195,142 @@ gchar *get_memory_total(void)
...
+        } else {
+            ret[rlen+1] = '\0';
+        }
+    }
+
+    fclose(dtmodel);
+    return ret;
+}
+
+gchar *rpi_get_boardname(void) {
```
lpereira: Make this `static` as well.

bp0:
I've made the requested changes.

Also, a new commit improves ARM processor information reporting. It now reports multiple cores with their min/max/current frequencies and defines several of the flags in the same manner as x86.

bp0:
I plan to split this into two different PRs.

lpereira:
There's no need to split. I'm away for the week on a company event, but
I'll be able to merge it this weekend.

Thanks for the patches!

bp0:
Alright then.

lpereira:
This is looking great, @bp0!

Issue #48, opened 2017-04-09
un-compilable due soup newer api


mckaygerhard:
since the commit 907a6249077745aa6d19eb7280823dd769223a29 i cannot compile.. 

the errors its due https://developer.gnome.org/libsoup/stable/libsoup-2.4-XMLRPC-Support.html#soup-xmlrpc-parse-response are only available since soup 2.52 

i understand due since 2.52 the `soup_xmlrpc_extract_method_response` are marked DEPRECATED in flavor of `soup_xmlrpc_parse_response` so the commit update the hardinfo compilation process to up to date soup api.. 

**maybe sound complicated but this made hardinfo unnavailable to compile for some**  that still uses the old soup api.. 

as i remenber some specific versions of hardware does not install with newer versions of software, as example VIA unichrome GPU does not offerts OpenGL capabilities with recent Xorg versions.. due not DRI v1 capabilities.. and yes i have a VIA motherboard and only works with some specific xorg version and specific linux kernel, some other examples are Asus Eeepc or Geode Daruma (yes also have one of that)

mckaygerhard:
the error:

```
libhardinfo-shell.a(syncmanager.c.o): In function `_soup_get_xmlrpc_value_int':
syncmanager.c:(.text+0x1a5): undefined reference to `soup_xmlrpc_parse_response'
libhardinfo-shell.a(syncmanager.c.o): In function `_soup_get_xmlrpc_value_string':
syncmanager.c:(.text+0x2c9): undefined reference to `soup_xmlrpc_parse_response'
libhardinfo-shell.a(syncmanager.c.o): In function `_soup_xmlrpc_call':
syncmanager.c:(.text+0x376): undefined reference to `soup_xmlrpc_message_new'
libhardinfo-shell.a(syncmanager.c.o): In function `_soup_xmlrpc_call_with_parameters':
syncmanager.c:(.text+0x473): undefined reference to `soup_xmlrpc_build_request'
libhardinfo-shell.a(syncmanager.c.o): In function `sync_dialog_start_sync':
syncmanager.c:(.text+0x970): undefined reference to `soup_session_new_with_options'
```

mckaygerhard:
umm after a investigation there's no api way compatible solution, must be a if-else conditional.. depending of the version detected must be used some calls.. but i not find how detect the version way of soup! the macro `SOUP_CHECK_VERSION` are only available since 2.42 and must compile agains 2.24, someone have some tips?
[...]

PR #64, opened 2017-05-23
cmake: add option to disable sync via libsoup

bp0:
Adds a cmake option HARDINFO_NOSYNC to disable libsoup
and remote sync.
* Workaround for incompatible libsoup (ex: on Raspbian) Issue #48
* xmlrpc.hardinfo.org is down and sync isn't currently
  available anyway.

Use: cmake -DHARDINFO_NOSYNC=1 ..

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 2.6)
 cmake_policy(VERSION 2.6)
 
 set(HARDINFO_VERSION "0.6-alpha")
-
+option(HARDINFO_NOSYNC "Disable database sync via libsoup" 0)
 SET( CMAKE_MODULE_PATH  "${CMAKE_CURRENT_SOURCE_DIR}/cmake" )
 include(GNUInstallDirs)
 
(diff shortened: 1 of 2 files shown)


mckaygerhard:
i do not have enought knowledge of cmake but its enought for the define of HAS_LIBSOUP macro?

bp0:
I'm, try it. It didn't work for me. The only way that I see is to disable searching for it at all, which is what this does.

mckaygerhard:
umm i have an idea but i'm too busy right now at my job.., there's one way that when cmake has the definition, added to cflags and cxxflags the -D as -DHAS_LIBSOUP and tri it ... 

take in consideration that i used older desktop and it seem work in some way some time ago..  due are very more faster rather the very slower modern linux version of today

bp0:
Running cmake will set HAS_LIBSOUP if libsoup is found, regardless of what you set. 

Doing it this way, as a cmake option, prevents cmake from even looking for the library, which prevents HAS_LIBSOUP from ever being set. It also has the advantage that it is visible as an option in cmake-gui with a description of what it does.

lpereira:
This can be made the default -- the hardinfo server is down in the meantime. There's no need to check for libsoup now.

Issue #60, opened 2017-04-30
Doesn't properly handle variable clock Skylake CPUs


justinkterry:
Current CPUs have min/max clock speeds that will be adjusted based on load. Benchmarking tools should always switch the processor to the highest clock speed (turbo) to properly determine performance. Using Ubuntu Gnome 17.04 on a Skylake series Intel i7-6700k CPU, this behavior does not occur. The CPU performance test all ran my CPU at the lowest possible clock speed. Please contact me if you have any questions or would like me to test something.

mckaygerhard:
the switch process in that CPU are based on load, and the real problem here its that the application does not represent a "necesary to stepping" to the CPU.. that due  the adoption of technologies that naturally prove speedier in certain applications. Yes, you get the expected notched-up chipset I/O throughput, as well as nods to more familiar technologies like DDR4 memory. This means that the CPU does not has spected behaviour.. and obviously its a windoze CPU due the intel drivers and utilities combinatios its the real stepping manager... as example the max base clock detecte was 127Mhz and the specifications said more, if a windoxe app reported that u spected, its due are hardcoded, trying olders version of same app reports differends numbers..

from my view, u have a fake CPU, that only works with specific things.. desktop Skylake doesn't present as compelling a story as we've seen from Intel's desktop chips in the past—another sign of how the chip world is rapidly moving away from the segment that has traditionally been its power king.

sorry for u, but the industry the only thing that wants its your money... the only CPU that hav a real ondeman its the quad-core 2.7GHz Core i7-6820HK CPU with an unlocked multiplier..

so in conclusion, application (hardinfo) does not have any control over the cpu, so application can handle the stepping change pver the CPU, and due Linux based OS are so good performance, that process does not represent a load charge.. this with that that CPU its made with specific apps (wuindoze like).. u dont have a right chooce... that u spected

inferno596:
But, at least user password authentication, it should be easily doable to switch the CPU into turbo for the tests. That's built into Linux already. Without that the bench marks aren't uniform or standardized between CPUs for bench marks as short as the ones included in hardinfo. 

Your claim that CPUs like this are windows exclusive is also unreasonable and completely wrong. High performance Intel CPUs dominate the server market, which is also dominated by Linux, hence Intel's massive open driver development team. Linux has excellent support for all modern Intel CPUs, and that's what the bulk of all the most recent kernel updates have concerned.

mckaygerhard:
[...]

PR #65, opened 2017-06-02
x86 improvements

bp0:
See individual commit messages for details.
This includes the patch from an older PR that improves the situation regarding issue #60.


modules/devices/x86/processor.c
```diff
@@ -610,6 +660,10 @@ gchar *processor_get_detailed_info(Processor * processor)
 			  "Frequency=%.2fMHz\n"
 			  "BogoMIPS=%.2f\n"
 			  "Byte Order=%s\n"
+			  "[Freq]\n"
```
lpereira: I'd prefer if this were not abbreviated: "Frequency", "Minimum", "Maximum", and "Current" are more consistent.

modules/devices/x86/processor.c
```diff
@@ -213,66 +214,111 @@ static void __cache_obtain_info(Processor *processor, gint processor_number)
     g_free(endpoint);
 }
 
+static gboolean _g_strv_contains(const gchar * const * strv, const gchar *str) {
```
lpereira: You can use `g_strv_contains()` in the code, and define this function using an `#if !GLIB_CHECK_VERSION(2,44)` block.  If this ends up being used by other parts of the code, it might be a good idea to put outside of a module and somewhere that can be reached by the project as a whole.

bp0:
The abbreviations have been expanded. I used "Frequency Scaling" instead of Frequency because there is already a Frequency item in the list that may be different from Maximum if cpuinfo reports a speed, but /sys/.../cpufreq/ isn't available.

Regarding _g_strv_contains(), I used the same function in the same way in modules/devices/arm/processor.c, but otherwise it doesn't appear anywhere else. 

1. Where do you think it should appear?
2. Which of these should it be? ( https://gist.github.com/bp0/ab4774dfc603664d8faedb4c2355e404 )

lpereira:
You can leave it in `hardinfo/util.c`. The first version is preferable. The prototype can live in `includes/hardinfo.h`.

PR #68, opened 2017-06-19
Build for AArch64, using the ARM configuration.

tsimonq2:
Ubuntu builds HardInfo for the arm64 architecture, and this patch is what we use to get it building. As far as I can tell, HardInfo works fine on that architecture.


CMakeLists.txt
```diff
@@ -29,6 +29,8 @@ elseif(${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "parisc.*")
 	set(HARDINFO_ARCH "parisc")
 elseif(${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "sparc{32,64}")
 	set(HARDINFO_ARCH "sparc")
+elseif(${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "aarch64*")
+	set(HARDINFO_ARCH "aarch64")
```
lpereira: Will it work better here if setting to "arm"? I don't have any aarch64 devices to test this, but there have been a few changes to the arm information gathering a few weeks ago, and it might work fine for aarch64.

tsimonq2: Within the next day or two I'll test that in the packaging on the Ubuntu infra. I'll let you know.

bp0: I've tested just using 'arm' on AArch64 (#72). It seems to work fine aside from what I think is an unrelated problem. I did notice that x86_64 uses symlinks to x86 instead, maybe it should be done the same way here.

lpereira:
Thanks for the patch -- I'm however merging @bp0's patch in #72, as it should give more information about aarch64 devices in the process.

tsimonq2:
wfm :+1:

PR #69, opened 2017-06-19
Make the strend function accessible

tsimonq2:
In Ubuntu, we had a problem building HardInfo on the s390x architecture because the way our build system is set up, it does not allow HardInfo to use the `strend` function if it was inlined. [Here](https://launchpadlibrarian.net/282109653/buildlog_ubuntu-yakkety-s390x.hardinfo_0.5.1-1.4ubuntu1_BUILDING.txt.gz) is the build log from that failure.

Making the changes in this pull request in Ubuntu's packaging made HardInfo buildable on s390x and caused no additional problems.

### DIFF
--- a/hardinfo/util.c
+++ b/hardinfo/util.c
@@ -113,7 +113,7 @@ gchar *seconds_to_string(unsigned int seconds)
     return g_strdup_printf(wdays, whours, wminutes);
 }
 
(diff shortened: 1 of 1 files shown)


lpereira:
Merged, thanks!

PR #72, opened 2017-06-20
ARM AArch64

bp0:
Like #68 except just use 'arm' for HARDINFO_ARCH when aarch64 as requested there. Note that this is different from x86_64 where symlinks to x86 are used instead.

Also, one /proc/cpuinfo will have architecture: 8, while another may have architecture: AArch64. An attempt is made to handle this with arm_arch() in arm_data.c.

There is a memory problem in populate_cpu_flags_list_internal(). If the printf is removed there will be a segfault. I will try to figure it out later, perhaps tomorrow.


bp0:
Seems to be working now.

lpereira:
Thanks for the patches!

PR #73, opened 2017-06-22
ARM socket/core and more cpufreq information

bp0:
Also, fix a code comment in x86.
screenshot_2017-06-22_16-28-04


lpereira:
Merged. Thanks!

mckaygerhard:
this code commits cause problems in x86:

errorenxeon

bp0:
It wouldn't be this one, it would be #65. Although if /sys/.../cpufreq/* isn't available, this is what it would look like; it will still use the the value from /proc/cpuinfo for frequency.

Issue #78, opened 2017-06-26
bug: no cpu socket/frecuency info with last x86 improvements


mckaygerhard:
i recently tested the hardinfo and on CPU x86 that are not scaling capable show 0 MHZ with recent changes:

![nofrecuency](https://user-images.githubusercontent.com/1571724/27540855-a199e7ae-5a4f-11e7-9d7a-8ae3b4cc8133.png)

does the code was tested before commit in x86 cpu's with no scale?

mckaygerhard:
seems does not work with any x86 cpu, here tested in a xeon in recent debian 8 up to date, same results in Devuan stable and squeeze:

![nofrecuency](https://user-images.githubusercontent.com/1571724/27541521-3440203a-5a52-11e7-9f10-af02b057aa3e.png)

bp0:
kernel version / distro, etc.
Is /sys/devices/system/cpu/cpu*/cpufreq/* existing and readable?

mckaygerhard:
kernel distro and version provided in issue.. the result of proc:

```
ls /sys/devices/system/cpu/cpu0/
cache  crash_notes  crash_notes_size  microcode  power	subsystem  topology  uevent
```

seems cpufreq related modules does not load due does not apply so in the code we need to put a conditional for those machines that does not support scaling

mckaygerhard:
seems apply too to you arm code improvement, so the conditional also must be apply there, due some early arm implementation does not have scaling features..

lpereira:
Closing issue as it's been fixed.

PR #80, opened 2017-06-26
arm,x86: Don't show frequency scaling information if it isn't available

bp0:
Hide Frequency Scaling section instead of showing zeros.
As requested in #78.


mckaygerhard:
hi @bp0 i see the commit, great but do you take care if also scale module are also loaded, in my shots there's no scaling module loaded...  so it maybe can be a problem?

lpereira:
Merged, thanks!

PR #82, opened 2017-07-06
Fix FTBFS on ppc64le

bp0:
Patch by Frédéric Bonnard <frediz@linux.vnet.ibm.com> from Debian bug report.
See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=865776#10

This is currently blocking migration.


lpereira:
Merged, thanks!

PR #83, opened 2017-07-06
PPC: Information improvements

bp0:
Based on the recent ARM work.

* Shows multiple processors, like x86 and ARM
* topology: physical package and core ids
* cpufreq: min/max/cur frequency and driver info

Tested against /proc/cpuinfos from a PowerMac G5 and the
Talos POWER8 workstation. Not tested on actual hardware.

### DIFF
--- a/includes/ppc/processor-platform.h
+++ b/includes/ppc/processor-platform.h
@@ -20,10 +20,18 @@
 #define __PROCESSOR_PLATFORM_H__
 
 struct _Processor {
+    gint id;
+    gchar *package_id, *core_id;
+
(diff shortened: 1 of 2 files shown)


mckaygerhard:
hi @bp0, either me, not have a ppc arch so i cannot test those improvement that for sure are great..  
but seems i can test in a mac/apple from 2013 with live disk of debian, that can work?

lpereira:
Thanks -- I'll test this on my Powerbook G4 later tonight.

bp0:
@mckaygerhard, if it doesn't have a ppc processor, then no.

@lpereira, right on.

lpereira:
It's much better on ppc, thanks!

Issue #38, opened 2016-12-27
Untranslatable strings


TotalCaesar659:
Hello! When I fixed Russian translation, I tried to add some untranslated strings to ru.po, converted to MO and launched the program with it. But the strings hasn't been translated, even if they was added in ru.po.
Here is the screenshots of some of that untranslatable places.
![2016-12-28-003048_1280x974_scrot](https://cloud.githubusercontent.com/assets/14265316/21510670/71b1ec8e-cca7-11e6-8d8a-22d9e999b79f.png)
![2016-12-28-003104_1280x974_scrot](https://cloud.githubusercontent.com/assets/14265316/21510673/76f0fd48-cca7-11e6-9e46-e64184b91c88.png)
![2016-12-28-003124_1280x974_scrot](https://cloud.githubusercontent.com/assets/14265316/21510675/7a6ab6b2-cca7-11e6-9cbd-1db1b1f1a3fd.png)
![2016-12-28-003144_1280x974_scrot](https://cloud.githubusercontent.com/assets/14265316/21510676/7d744170-cca7-11e6-903e-7efde17a2bff.png)

lpereira:
Nice finds. I'll take a look later.

mckaygerhard:
hi @TotalCaesar659 , please can u confirm that the memory amount are displayed correctly, in computer->Resume->Memory (please revise it in russian env, not in english) to confirm a bug in my translated po file or in code

TotalCaesar659:
Hi @mckaygerhard, I don't see "Resume" in program. Which point I have to check on these pictures?
![2016-12-28-221951_1280x974_scrot](https://cloud.githubusercontent.com/assets/14265316/21529789/29cef1b8-cd4d-11e6-967c-7ce42bfae860.png)
![2016-12-28-222104_1280x974_scrot](https://cloud.githubusercontent.com/assets/14265316/21529794/31cbf49c-cd4d-11e6-86f4-805122f7b276.png)
I can reboot system to Russian and send a screenshot.

mckaygerhard:
sorry my fault, in Computer->summary->memory , in spanish i cannot see the amount but in english amount of memory appears! can u confirm if in russnik appears correctly the translated stringa nd also the amount? also double-chek in devices->memory too, send screenshot if u can but its optionally

mi interes its that if the string memory are translated and shows amount of ram!

TotalCaesar659:
Here is the screenshot in Russian.
[...]

Issue #31, opened 2016-02-16
memory info not show with poor translated env


mckaygerhard:
in the resumen module in computer section memory info its not show if hardinfo are used in a poorly translated language.

as @lpereira said:

> If it's not updating, it's a bug somewhere else, likely in the shell rather than the computer module. I find it very unlikely, as that is some well-tested part of the code and it works on every machine I've tried HardInfo on.

added 

> it's possible that it's been poorly translated and the UpdateInterval$Memory=1000 has not been translated to match the Memory key.

then i tested as and have right in part, using different lang this king of memory info that show are broken, i tested in russian and in spanish and does not show nothing... 

using this for launch:

```
export LANG=C;export LANGUAGE=C;export GDM_LANG=C;export LC_ALL=C;hardinfo
```

hardinfo correctly show so i'll open a issue and let this pull request opened until u decided what are best to do

mckaygerhard:
firts of all i'll revise the translation po fiels and tried to complete most possible and then post feedback here

mckaygerhard:
related to pull request #28

mckaygerhard:
i research and found its well translated the es.po file:

i have strings:

"UpdateInterval$Memory=1000\n"
and
"UpdateInterval$Memoria=1000\n"

so i dont knwon what other .. i research more in few days..

mckaygerhard:
i confirmed in the isue comment that the pool request mention, see #28, the pull request its not elegant but works for now, quoted :

> Look at the Memory, it can't be translated, because the value disappears with it. There is the comment "//Load Average is not updated if locale is not C, switch locale to C" in /modules/computer.c (https://github.com/lpereira/hardinfo/blob/master/modules/computer.c).

mckaygerhard:
the commits from pull #86 solves other issues like  #38 and this! see picture: i can close when the pull are merged!

![hardinfomemetranslatable](https://user-images.githubusercontent.com/1571724/28135497-6186c4d0-6714-11e7-81f4-42018031c774.png)

mckaygerhard:
this issue was solved, closed with pull #36 merged

PR #86, opened 2017-07-08
Make processor info strings translatable

bp0:
Break up the big sprintf() format string for processor (x86) that was translated as one piece into smaller more re-usable pieces. 
* the recently added topology and cpufreq information labels can be translated
* common strings like "Processor", "Capabilities", "Frequency", etc
* "Little Endian" and "Big Endian"
* some unit labels "kHz", "MHz", "ns" (maybe not so useful ... or maybe so, it seems that Russian MHz is different)

Only tested on x86 and ARM hardware so far. For all others, /proc/cpuinfo reading has been tested, but not on actual hardware.

Update:
* modules/devices/*/processor.c are all cleaned up. All strings are translatable.
* Added CMake options and a test script to test that code without the actual hardware.
* all processor.c are tested and working
* riscv added
* ia64 improved
* parisc repaired and improved


mckaygerhard:
after that pull i can help with the missing string on spanish and some russain

bp0:
@mckaygerhard, right on.

modules/devices/ia64/processor.c
```diff
@@ -19,36 +19,48 @@
 #include "hardinfo.h"
 #include "devices.h"
 
+gchar *byte_order_str() {
```
lpereira: This can be moved to somewhere common.

modules/devices/ia64/processor.c
```diff
@@ -19,36 +19,48 @@
...
+#if G_BYTE_ORDER == G_LITTLE_ENDIAN
+    return _("Little Endian");
+#else
+    return _("Big Endian");
+#endif
+}
+
+#ifndef PROC_CPUINFO
+#define PROC_CPUINFO "/proc/cpuinfo"
```
lpereira: Same for this.

lpereira:
Excellent, thanks! Just a few minor nitpicks, though.

bp0:
I'm done for now. This could be merged, if you like.

bp0:
Regarding the flag definitions, it really seems like there needs to be some additional info for the translator.
There are a couple ways that might be done.

Use code comments. 
* Adjust xgettext params, like `xgettext -ctrans ...`
* Add comments in the code:
> ...
>  `{ "fpu",	N_(/* trans: x86 flag fpu*/ "Onboard FPU (floating point support)") },` 
> ...
* Result in the .po:
> #. trans: x86 flag fpu
> #: modules//devices/x86/processor.c:339
> msgid "Onboard FPU (floating point support)"
> msgstr ""`

Alternately, gettext context could be provided, using pgettext(), but this is a little more complicated when it comes to calling gettext. See [gettext contexts](https://www.gnu.org/software/gettext/manual/html_node/Contexts.html)

bp0:
Regarding using gettext contexts and comments for translatable flag definitions,
It looks like this:
* https://github.com/bp0/rpiz/blob/trans/src/arm_data.c#L42
* https://github.com/bp0/rpiz/blob/trans/src/riscv_data.c#L38

A snippet of the .po(t) result:
>#. /ext:vfpv4
>#: arm_data.c:58
>msgctxt "arm-flag"
>msgid "VFP version 4 with fast context switching"
>msgstr ""

context arm-flag, and comment ext:vfpv4 so it is known what the string is supposed to define.

bp0:
The problems with the special fields "..." are resolved. Solves the "unknown field" portion of #38.

mckaygerhard:
this pull request must be renamed to "translatable strings" due solves other issues like  #38 and #31 
please merge and then close that issues! a picture to demostrate (non tranlated its due are missing in po file)
hardinfomemetranslatable

Obliteron:
> I suggest uploading the source code of the language in Transifex, so everyone can contribute to the translation!
> Eu sugiro fazer o upload do código-fonte do idioma no Transifex, assim todo mundo pode contribuir para a tradução!
@lpereira

Obliteron:
>>> I suggest uploading the source code of the language in Transifex, so everyone can contribute to the translation!

>>> Eu sugiro fazer o upload do código-fonte do idioma no Transifex, assim todo mundo pode contribuir para a tradução!
@lpereira

mckaygerhard:
well github its already a colaboration tool, and the po files inside the repository can be handle locally by linux po trasnaltion tools, one the git repo are cloned, so the tranfiex "layer" can complicated simple things...

bp0:
Oh, that last one conflicts with the fix for libc detection. I was trying to avoid that, but I messed up. I'll stop now until this set is merged.

lpereira:
Merged, thank you very much!

PR #88, opened 2017-07-13
updatepo.sh: sort output of find, use more logical search order

bp0:
An attempt to make updatepo.sh generate hardinfo.pot and .po files
that are the same regardless of the machine, so that diffs are less
noisy and more useful.

The second commit is just a run of the updated updatepo.sh.

Third is a fix for #89, and a few other tweaks described in the commit message.


bp0:
> that line are a extra line.. that i think must removed

python2 -v returns the version in stderr, python3 -v returns the version in stdout. That's why there are multiple lines. I don't know if `python` will always point to python2 (what you say about Debian 7 vs 8 seems to suggest not), so I included that line too. Maybe one of (python,python2) can be removed.

> another example, theres php5 and php7 that have different behaviour, also ruby 1.8 and ruby 1.9 and you do not put same lines for!

I don't know about different php versions, or ruby. I just edited the part I know something about. Someone else can edit those if they are wrong or if more need to be added.

> i see that like king of racis .. too much python.. in linux, and its not secret that the famous of are due intel and MS system compatibility..

I don't follow.

mckaygerhard:
> python2 -v returns the version in stderr, python3 -v returns the version in stdout. That's why there are multiple lines. 
there's more than a qt versions installed in some linux, but not various lines in hardinfo and too much focusin/importance in python.. but not in gcc or php, a image to ilustrate better that i said: 

> I don't know if python will always point to python2 (what you say about Debian 7 vs 8 seems to suggest not),
see image

variousversionpythonwhy

i have qt4 and qt5, but hardinfo only show info of one!

i have py 2.7 and py 3.X and hardinfo show version of many python..

mckaygerhard:
in short:
i have qt3 (trynity desktop), qt4 and qt5, but hardinfo only show info of one!
i have php5 and php7 but hardinfo only show info of one
i have gcc2.x, gcc3.4 (oracle need it), and gcc 4.X but hardinfo only show info of one
i have py 2.7 and py 3.X and hardinfo show version of !many python!

bp0:
Python 2 and 3 are different languages. Same with perl 5 and 6.

mckaygerhard:
ajah! and hardinfo only show info of one perl! same for qt! that's my point

NOTE: different api sintax, not different languajes...

bp0:
> and hardinfo only show info of one perl!

Same patch that adds python3 adds perl6:
https://github.com/lpereira/hardinfo/pull/88/files#diff-264bfb8536f8084d8c52d47b594b891fR206

More could be added.

mckaygerhard:
;-) ok, we need two: for gambas2 and gambas3 due gambas its the most easy to learn must be there!
also for gcc, php5 vs php7, php4  X-D
so its nonsense to added many lines as "sintax types" wil have a languaje! uff

so (as python author cited) python 2 not was good designed, now python3 are well designed and future python4, python5 and in future hardinfo will have a section only to python

lpereira:
@bp0 Why was this closed?

bp0:
It completely breaks translation as it was. I'll have a new series soon.

On Jul 14, 2017 8:36 PM, "Leandro A. F. Pereira" <notifications@github.com>
wrote:

@bp0 <https://github.com/bp0> Why was this closed?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/lpereira/hardinfo/pull/88#issuecomment-315500785>, or mute
the thread

PR #92, opened 2017-07-15
Translation support fixes, device tree support, x86 bug detection fix, others

bp0:
Mostly described in commit messages.

For "Make locale dir match install path", the default install location is /usr/local/  but the hard-coded messages location was "/usr/share/locale". This bug was a bit frustrating to find.

PR #93, opened 2017-07-17
Add "Predictable Network Names" from systemd

bp0:
See:
https://github.com/systemd/systemd/blob/master/src/udev/udev-builtin-net_id.c#L20
and
https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/


bp0:
I've also made all the net.c strings translatable, but it depends on #92 because it uses contexts.

bp0:
Needs #92 or it will break.

lpereira:
Merged, thanks!

PR #97, opened 2017-07-19
Device Tree related changes

bp0:
* Added some Device Tree .dts to test/data. Please add more. Use `dtc -I fs /proc/device-tree` or perhaps `dtc -I fs /sys/firmware/devicetree/base` to create.
* Fix bug where every property appeared as byte sequence sometimes.
* Look for Device Tree root in two known locations. Please add more if you find it elsewhere.
* Decode /_ _ overrides _ _ fields
* Started some work decoding clocks and gpios, currently turned off because it is incomplete.
* reg grouping using inherited #address-cells and #size-cells.


lpereira:
Merged, thanks!

PR #98, opened 2017-07-20
Make all column headers translatable; break out of c-format strings

bp0:
I think the labels of column headers and/or rows should be broken out of
the c-format strings because:
* The strings are more easily translated.
* The strings are more likely to be re-usable.
* A small change to one does not lose all of them.
* Changing a shell parameter does not lose all of them.
* Re-ordering them does not lose all of them.
* Adding items does not lose all of them.

I've been doing this in all new code, and converting old code.

Example:

	return g_strdup_printf(_("[$ShellParam$]\n"
			   "ReloadInterval=10000\n"
			   "ColumnTitle$TextValue=Name\n"
			   "ColumnTitle$Value=Group ID\n"
			   "ShowColumnHeaders=true\n"
			   "[Groups]\n"
			   "%s\n"), groups);

Changed to:

	return g_strdup_printf("[$ShellParam$]\n"
	               "ReloadInterval=10000\n"
	               "ColumnTitle$TextValue=%s\n" /* Name */
	               "ColumnTitle$Value=%s\n" /* Group ID */
	               "ShowColumnHeaders=true\n"
	               "[%s]\n%s\n",
	               _("Name"), _("Group ID"),
	               _("Groups"), groups);

Another example (post-conversion):

	gchar *val = g_strdup_printf("[%s]\n"
	        "%s=%d\n"
	        "%s=%d\n"
	        "%s=%s\n"
	        "%s=%s\n",
	        _("User Information"),
	        _("User ID"), (gint) passwd_->pw_uid,
	        _("Group ID"), (gint) passwd_->pw_gid,
	        _("Home Directory"), passwd_->pw_dir,
	        _("Default Shell"), passwd_->pw_shell);


bp0:
Progress tracked at #100. This PR can be merged any time, if you like.

bp0:
I rebased to master, updated the .pot/.po files and restored any missing translated labels. This should be merged soon, or it will be a messy conflict. 
When I get back to working on this, I'll save the labels, as I did here.

lpereira:
The modules sending back strings to the shell was one of the worst decisions I've made in HardInfo.  While this effectively solves a problem, I wonder if we can't do better.  For instance, what if we had something that allowed the following?

```c

    Info *info = info_new();

    info_set_shell_params(info, (struct InfoShellParam) {
        .view_type = ...,
        .sort_type = ...,
    });

    info_add_group(info, _("Version"),
        info_field(_("Kernel"), computer->os->kernel),
        info_field(_("Version"), computer->os->kernel_version),
        info_field(_("C Library"), computer->os->libc),
#if defined(FOO)
        info_field(_("Distribution"), computer->os->distro),
#endif
        NULL);

    info_add_group(info, _("Sensors"),
        info_field_params(info_field("CPU Temp0", value[0]),
            (struct InfoParam) { .icon = "temperature.png", .update_interval = 1000 });
        info_field_params(info_field("CPU Fan0", value[0]),
            (struct InfoParam) { .icon = "fan.png", .update_interval = 1000 });
        info_field_params(info_field("CPU Voltage0", value[0]),
            (struct InfoParam) { .icon = "voltage.png", .update_interval = 1000 });
        NULL);

    return info_flatten(
        info_add_group(...),
        info_add_group(...),
        info_add_group(...),
        NULL);
```

There are some advantages here:

- ShellParam order wouldn't matter any longer
- Field parameters wouldn't require changes to ShellParam
- #ifdefs can be used easily
- It retains the translation abilities from this patch
- It removes problems with escaping and other issues that plagued the current approach for years
- The `info_flatten()` function at this stage would be something that would create the same string we generate to not change the shell at this point. When this is mature, the shell can be modified to work with this directly
- Other field types can be easily added (for lines with multiple columns, for instance)

What do you think about this?

bp0:
I think it's a good idea and I started to work on a similar thing, with the
additional goal of adding all the data to a single large data structure as
it is scanned, where the data can be searched for and values that can be
pinned onto a graph, even when running the benchmarks.

This seems like all that is a couple minutes from working though, so for
now I think these patches are a good way to move forward.


On Jul 30, 2017 11:24 AM, "Leandro A. F. Pereira" <notifications@github.com>
wrote:

The modules sending back strings to the shell was one of the worst
decisions I've made in HardInfo. While this effectively solves a problem, I
wonder if we can't do better. For instance, what if we had something that
allowed the following?

    Info *info = info_new();

    info_set_shell_params(info, (struct InfoShellParam) {
        .view_type = ...,
        .sort_type = ...,
    });

    info_add_group(info, _("Version"),
        info_field(_("Kernel"), computer->os->kernel),
        info_field(_("Version"), computer->os->kernel_version),
        info_field(_("C Library"), computer->os->libc),
#if defined(FOO)
        info_field(_("Distribution"), computer->os->distro),
#endif
        NULL);

    info_add_group(info, _("Sensors"),
        info_field_params(info_field("CPU Temp0", value[0]),
            (struct InfoParam) { .icon = "temperature.png",
.update_interval = 1000 });
        info_field_params(info_field("CPU Fan0", value[0]),
            (struct InfoParam) { .icon = "fan.png", .update_interval = 1000 });
        info_field_params(info_field("CPU Voltage0", value[0]),
            (struct InfoParam) { .icon = "voltage.png",
.update_interval = 1000 });
        NULL);

    return info_flatten(
        info_add_group(...),
        info_add_group(...),
        info_add_group(...),
        NULL);

There are some advantages here:

   - ShellParam order wouldn't matter any longer
   - Field parameters wouldn't require changes to ShellParam
   - #ifdefs can be used easily
   - It retains the translation abilities from this patch
   - It removes problems with escaping and other issues that plagued the
   current approach for years
   - The info_flatten() function at this stage would be something that
   would create the same string we generate to not change the shell at this
   point. When this is mature, the shell can be modified to work with this
   directly
   - Other field types can be easily added (for lines with multiple
   columns, for instance)

What do you think about this?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/lpereira/hardinfo/pull/98#issuecomment-318912577>, or mute
the thread

lpereira:
Sure. I'll merge these. Thanks again!

PR #104, opened 2017-07-20
Update README to corrected information

mckaygerhard:
also reflect a format markdown for better
* added hardinfo status information event a header with "want a mantainer"
* update build/requirement information with markdwon format
* added memory speed necesary info missing
* corrrected the sensor tip, due are related only to debian..


lpereira:
Merged, thanks

Issue #79, opened 2017-06-26
bug: no glibc available info now


mckaygerhard:
i cannot track from where commit the libc6 info cannot be tracket.. but now its unkown, i tested in Debian jeesie, Devuan jeesie, Debian wheeze, and only works in Debian squeeze and lenny works too perfectly, see picture:

![errorcglibchardinfo](https://user-images.githubusercontent.com/1571724/27542667-f35d976a-5a55-11e7-8917-43efad1e1938.png)

Debian in recent packages has a patch but its unusable (was using ldconfig -V and strip the version info) due neither works.

mckaygerhard:
tested in linux x86 embebed (daruma) and works, its equivalent with debian squeeze, due recent stable versions are too slower and vendor does not want to update..

mckaygerhard:
seems the problem its related only to Debian, due in alpine works..

mckaygerhard:
ok, the top patch https://developer.gnome.org/glib/stable/glib-Spawning-Processes.html#g-spawn-command-line-async seem both calls are glib older compatible.. i'll test in other later in night and report feedback

mckaygerhard:
tested in various hardwares with debian squeeze, worked:

**debian older << 7 , squeeze, works:** tested in dell r15, dell powered, dell t110 and asus eepc with debian squeeze 6.0, 
**debian newers >> 7 (jessie, sid), dont:** tested in debian jeesie not working, see image, ill test now in devuan not working neither.. i do not tested in other distros rather debian derived!

see images:

![hardinfo-error-gettingglibcerror](https://user-images.githubusercontent.com/1571724/28135096-1535651a-6713-11e7-821d-b159b9d46e37.png)

![hardinfo-glibc6-debiansqueeze](https://user-images.githubusercontent.com/1571724/28135106-1fd3776e-6713-11e7-8e42-c0dec212892c.png)

bp0:
The format string changed in the fix. It no longer has the (stable) part, so you're testing a version of hardinfo that doesn't have the recent fixes to libc detection.

mckaygerhard:
hi @bp0 i tested with your brand, and also with the last brand of stable at yesterday.. and now i tested today with the lasted in git.. and same results..

i not tested in debian wheeze yet.. i think its a problem in the way of glibc returns the version strings.. i not have another distro to test .. but i could install today a winbuntu and test.. , i'll make it with kubuntu-trinity and report feedback...

mckaygerhard:
[...]

PR #107, opened 2017-07-21
Another libc detection fix

bp0:
Starting in Debian 9, ldconfig requires root, and so it can't be
used to get the GNU libc version. Try `ldd --version` first. It should
give the same result. (#79)


lpereira:
Merged, thanks!

Issue #101, opened 2017-07-20
Fix all issues that Debian or Ubuntu carry patches for


bp0:
https://anonscm.debian.org/git/collab-maint/hardinfo.git/tree/debian/patches
- [ ] make-build-reproducable.patch ?
- [x] fix-ppc64el-ftbfs.patch 
- [x] glibc-version.patch
- [ ] man page

bp0:
@tsimonq2, there are better fixes here for ppc support and glib detection than are provided by those patches.

tsimonq2:
@bp0 So you're saying that if I update to the latest Git snapshot in my packaging, I can drop `fix-ppc64el-ftbfs.patch` and `glibc-version.patch`?

As for the reproducibility patch, that's still a WIP and we're working with the Debian Reproducible Team to discuss the best route.

bp0:
@tsimonq2, I am saying that. 
Although, pull/wait for #97 first. :smile: And maybe #98, #99...

mckaygerhard:
either the  glibc-version.patch or the current detection on the git work today on debian, both process faild as in #79 see pictures .. recent commits of today make some difference?

bp0:
The fixes were merged about a week ago, 10 July.
As I already mentioned, (https://github.com/lpereira/hardinfo/issues/79#issuecomment-314980276) your images show that you are not using a snapshot with the fixes applied.

mckaygerhard:
ummm, i was using, but i'll retest and compile agains git right now! i compiled agains recent git! what i should do bad them?

![2017-07-20-105912_1366x768_scrot](https://user-images.githubusercontent.com/1571724/28424009-90c6712e-6d3a-11e7-8a73-30d45743872c.png)

mckaygerhard:
the snapshot previous was agains a recent installation clean, the hardinfo was the firts compilation made in, its a recent machine

bp0:
Anyway, the fix here is just an enhanced and more robust version of the Debian patch (use the first line of `ldconfig -V`, skip the first word). If it doesn't work with hardinfo master, it doesn't work with the Debian patch either. The Debian patch could still be dropped, and should be.

mckaygerhard:
yeah the fix in git are a improved version of debian, so i tested that works in olders version of debian i tested in Debian lenny and debian squeeze,  all older versions of debian.. only fails on recent versions

mckaygerhard:
[...]

PR #109, opened 2017-07-22
Add cmake option to make a "Reproducible Build"

bp0:
For Debian, see #101.
Leave out the build kernel version so that the build will be byte-exact, built from any machine.
Usage: cmake -DREPRODUCIBLE_BUILD=1


lpereira:
I don't think we need an option for this; this could be the default and the only option.

bp0:
Done.

lpereira:
Merged, thanks!

PR #110, opened 2017-07-23
Add Power Mac data from cpuinfo and device tree to summary

bp0:
Also fixes an old bug in computer/memory.c that caused warnings at run time.


lpereira:
Merged, thanks

modules/devices/devicetree/pmac_data.c
```diff
@@ -0,0 +1,99 @@
+/*
+ *    HardInfo - Displays System Information
+ *    Copyright (C) 2003-2007 Leandro A. F. Pereira <leandro@hardinfo.org>
```
lpereira: For new files, you can use your name here. If you think you've done extensive changes to the file and would also like to be the copyright holder, you can add your name to other files as you touch them.

PR #111, opened 2017-07-23
Reset sort order on set_view_type()

### LINKED ISSUE #16: Reversed polarity for progress bars
The progress bar indicator is reversed

Here is a screenshot
http://s4.postimg.org/o19x16p65/kde_polarity.png

bp0:
Required to fix #16, for me.

### DIFF
--- a/shell/shell.c
+++ b/shell/shell.c
@@ -897,6 +897,7 @@ static void set_view_type(ShellViewType viewtype, gboolean reload)
 
     shell->normalize_percentage = TRUE;
     shell->view_type = viewtype;
+    shell->_order_type = 0;
 
     /* use an unsorted tree model */
     GtkTreeSortable *sortable = GTK_TREE_SORTABLE(shell->info->model);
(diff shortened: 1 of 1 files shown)


shell/shell.c
```diff
@@ -897,6 +897,7 @@ static void set_view_type(ShellViewType viewtype, gboolean reload)
 
     shell->normalize_percentage = TRUE;
     shell->view_type = viewtype;
+    shell->_order_type = 0;
```
lpereira: I'm curious why this fixes the issue: this is setting the order type to descending.

bp0: The % is inverted later if shell->_order_type = 1.
https://github.com/bp0/hardinfo/blob/2e585991f0880a8a6596c65a364c622d28b24961/shell/shell.c#L1245

bp0: The benchmarks each specifically set OrderType in the shell params. Filesystems does not, so it ends up using whatever the last one was. Resetting to 0 (descending) and having that as a default seems like the right fix, considering that 1 (ascending) has that special handling.

lpereira: Alright, makes sense. Please just use the enum value instead of 0 for clarity.

lpereira:
Squashed and merged, thanks! (For small fixups such as these, please use interactive rebase, squash/fixup, and force push to the branch the PR has been opened. This makes my life a little bit easier and reduces noise in the git history.)

PR #113, opened 2017-07-25
GTK3

bp0:
v2: #113.
-----
Currently, this is the same as #59, but rebased with conflicts resolved. I'll push my changes here. 
TODO at #112.

So far:
* Added cmake option for gtk3
* hardinfo/util.c should be working, but deprecation warnings remain for gtk_ui_manager_remove_ui() and GtkStock.
* shell/report.c should be working but a deprecation warning remains for gtk_dialog_get_action_area().
* builds and runs, but has bugs (see https://github.com/lpereira/hardinfo/issues/112#issuecomment-317666963)
* loadgraph doesn't work

Note: "test555" is @TotalCaesar659. That's how it appears in the commit at #59. Perhaps if ever this gets merged, pause after that commit and `commit --amend --author="TotalCaesar659 <x24cm5b8c54q6szxw@yandex.ru>"` to correct the author.

PR #59, opened 2017-04-20
Added GTK3 support (unfinished, need help)

TotalCaesar659:
OK, here is my sketches of adding GTK3 support to HardInfo. After compiling with gtk3, there are error messages something like this:

>/home/maxpayne/hardinfo/shell/shell.c:841:51: error: ‘GtkScrolledWindow {aka struct _GtkScrolledWindow}’ has no member named ‘vscrollbar’
          (GTK_SCROLLED_WINDOW(shell->tree->scroll)-> \
                                                   ^
/usr/include/glib-2.0/gobject/gtype.h:2207:57: note: in definition of macro ‘_G_TYPE_CIC’
     ((ct*) g_type_check_instance_cast ((GTypeInstance*) ip, gt))
                                                         ^
/usr/include/gtk-3.0/gtk/gtkrange.h:40:36: note: in expansion of macro ‘G_TYPE_CHECK_INSTANCE_CAST’
 #define GTK_RANGE(obj)            (G_TYPE_CHECK_INSTANCE_CAST ((obj), GTK_TYPE_RANGE, GtkRange))
                                    ^
/home/maxpayne/hardinfo/shell/shell.c:840:26: note: in expansion of macro ‘GTK_RANGE’
      gtk_range_get_value(GTK_RANGE \
                          ^
/home/maxpayne/hardinfo/shell/shell.c:856:23: note: in expansion of macro ‘RANGE_GET_VALUE’
     pos_info_scroll = RANGE_GET_VALUE(info, vscrollbar);
                       ^

I don't know how to solve them, so I hope that it may be some starting point for further implementing of GTK3.

PR #114, opened 2017-07-25
GTK3 Support [v2]

bp0:
This PR replaces #113 (and #59). It corrects the Author on @TotalCaesar659's work and squashes all of my commits into one.

To test:
```
mkdir build && cd build
cmake -DHARDINFO_GTK3=1 ..
[etc.]
```

Hardinfo builds and runs without crashing on GTK3.
* there are many deprecation warnings during build.
* loadgraph is replaced by Christian Hergert's uber-graph widget for GTK3 hardinfo
* guibench is disabled for gtk3.
* gtk2 continues to work as well as it ever did.
* look for TODO:GTK3 in the code to find places that need work.


CMakeLists.txt
```diff
@@ -176,31 +193,70 @@ foreach (_module ${HARDINFO_MODULES})
...
 
-add_library(hardinfo-shell STATIC
-        shell/callbacks.c
-        shell/iconcache.c
-        shell/loadgraph.c
-        shell/menu.c
-        shell/report.c
-        shell/shell.c
-        shell/stock.c
+if (HARDINFO_GTK3)
```
lpereira: Can uber-graph be used in GTK2? If so, we can just get rid of the old loadgraph code and use it instead.

This should simplify the build system as well by removing some conditionals and copied lists of files.

hardinfo/util.c
```diff
@@ -979,6 +994,9 @@ gint tree_view_get_visible_height(GtkTreeView * tv)
 
 void tree_view_save_image(gchar * filename)
```
lpereira: This whole function can be removed, it's not used for anything. (It's a leftover from a feature I've never finished.)

includes/loadgraph.h
```diff
@@ -39,37 +33,20 @@ typedef enum {
...
-    gint	   width, height;
-    LoadGraphColor color;    
-    
-    gint	   max_value, remax_count;
-    
-    PangoLayout   *layout;
-    gchar	  *suffix;
-};
+void         load_graph_update(LoadGraph *lg, gdouble value);
+void         load_graph_update_ex(LoadGraph *lg, guint line, gdouble value);
```
lpereira: There are very few places where `load_graph_update()` is called. Maybe it would be better to make `load_graph_update_ex()` the default and only behavior to update the load graph.

shell/callbacks.c
```diff
@@ -44,13 +44,23 @@ void cb_save_graphic()
     gchar *filename;
```
lpereira: The `cb_save_graphic()` function can be removed as well, nothing calls it any longer.

shell/loadgraph.c
```diff
@@ -20,11 +20,38 @@
...
  * License along with the Simple Load Graph; if not, write to the Free
  * Software Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA
- * 02111-1307 USA. 
+ * 02111-1307 USA.
  */
 
 #include "loadgraph.h"
 
+struct _LoadGraph {
+#if GTK_CHECK_VERSION(3, 0, 0)
```
lpereira: Is this change necessary if uber-graph is being used for GTK3 now?

shell/report.c
```diff
@@ -636,15 +646,21 @@ static gboolean report_generate(ReportDialog * rd)
 					GTK_MESSAGE_QUESTION,
 					GTK_BUTTONS_NONE,
 					_("Open the report with your web browser?"));
+#if GTK_CHECK_VERSION(3, 0, 0)
+    gtk_dialog_add_buttons(GTK_DIALOG(dialog),
+			       "_No", GTK_RESPONSE_REJECT,
```
lpereira: Shouldn't these be translatable?

shell/shell.c
```diff
@@ -33,6 +33,10 @@
 
 #include "callbacks.h"
 
+#ifndef gtk_notebook_set_page
```
lpereira: Better to use the `GTK_CHECK_VERSION` macro here.

CMakeLists.txt
bp0: uber-graph requires gtk3.

lpereira:
Merged, thanks @bp0 and @TotalCaesar659!

PR #115, opened 2017-07-25
device tree: fix crash when device tree is not found

bp0:
* check if device tree was found before add_keys()
* UNKIFNULL(model) before strcmp
* make UNKIFNULL() STRIFNULL() EMPIFNULL() macros more generic


lpereira:
Merged, thanks!

Issue #118, opened 2017-07-28
Any objection to moving the gtk2 requirement from 2.10 to 2.18?


bp0:
It would remove all of the GTK_CHECK_VERSION()s for 2.x, and many of the changes for GTK3 wouldn't require version checks because they were existing by 2.18, but the 2.x series still carried the functions that were replaced for compatibility. The compatibility was only removed in 3.0.
GTK+2.18 was released around September 2009.

After PR #125:
```
grep -nr --include='*.[hc]' "GTK_CHECK_VERSION(2" *
shell/report.c:805:#if GTK_CHECK_VERSION(2, 14, 0)
shell/report.c:896:#if GTK_CHECK_VERSION(2, 18, 0)
shell/report.c:907:#if GTK_CHECK_VERSION(2, 18, 0)
shell/report.c:915:#if GTK_CHECK_VERSION(2, 14, 0)
shell/report.c:933:#if GTK_CHECK_VERSION(2, 18, 0)
shell/report.c:943:#if GTK_CHECK_VERSION(2, 18, 0)
shell/shell.c:123:#if GTK_CHECK_VERSION(2,16,0)
shell/shell.c:515:#if GTK_CHECK_VERSION(2, 18, 0)
shell/shell.c:843:#if GTK_CHECK_VERSION(2, 14, 0)
shell/shell.c:862:#if GTK_CHECK_VERSION(2, 14, 0)
shell/shell.c:896:#if GTK_CHECK_VERSION(2, 14, 0)
shell/shell.c:952:#if GTK_CHECK_VERSION(2, 18, 0)
shell/shell.c:1001:#if GTK_CHECK_VERSION(2, 18, 0)
shell/shell.c:1017:#if GTK_CHECK_VERSION(2, 18, 0)
shell/shell.c:1401:#if GTK_CHECK_VERSION(2, 14, 0)
shell/shell.c:1409:#if GTK_CHECK_VERSION(2, 14, 0)
shell/shell.c:1464:#if GTK_CHECK_VERSION(2, 14, 0)
shell/shell.c:1475:#if GTK_CHECK_VERSION(2,12,0)
shell/shell.c:1963:#if GTK_CHECK_VERSION(2,12,0)
shell/syncmanager.c:670:#if GTK_CHECK_VERSION(2, 14, 0)
shell/syncmanager.c:739:#if GTK_CHECK_VERSION(2, 14, 0)
[...]

PR #119, opened 2017-07-31
Various recent stuff

bp0:
Changes based on recent merges and discussions.

* GTK_CHECK_VERSION()-related: shows some code that doesn't need gtk3. If moving up to 2.18 as a requirement (#118), these and all existing conditional code segments for 2.x could be removed. 
* loadgraph gtk2 only, doesn't need conditionals for gtk3.
* Some recent changes to comptuer/os.c caused some compiler warnings.
* Removing tree_view_save_image(), a feature that was "never finished" (https://github.com/lpereira/hardinfo/pull/114#discussion_r129861059)
* The fix to network statistics is just something that was waiting to be in a PR somewhere.

### DIFF
--- a/modules/network.c
+++ b/modules/network.c
@@ -80,6 +80,7 @@ void scan_statistics(gboolean reload)
     FILE *netstat;
     gchar buffer[256];
     gchar *netstat_path;
+    int line = 0;
 
     SCAN_START();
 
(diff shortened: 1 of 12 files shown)


modules/network.c
```diff
@@ -99,22 +100,16 @@ void scan_statistics(gboolean reload)
...
-            *tmp2 = 0;
-            tmp2++;
+          } else {
+            gchar *tmp = buffer;
 
-            *tmp2 = toupper(*tmp2);
+            while (*tmp && isspace(*tmp)) tmp++;
 
-            __statistics = h_strdup_cprintf("%s=%s\n",
+            __statistics = h_strdup_cprintf("<b> </b>#%d=%s\n",
```
lpereira: Isn't this missing something? Why is `<b> </b>` here?

bp0: It's a hack, but it works. The keys must be unique for each line, and everything after # is hidden. Something must be before the # or the key is "" after the spaces are stripped and it won't be displayed. A bolded space won't be stripped.

bp0: Network statistics wasn't really working at all before this patch. Since the stat values were being used for the keys, and keys must be unique. If two stats had the same value they would also both have the same text.

For example:
* 2 outgoing packets dropped
* 2 packets reassembled ok

... becomes:
* 2 outgoing packets dropped
* 2 outgoing packets dropped

Also, the stats that weren't in the form `%d %s` were ignored. This "fix" basically just puts each line of output of from `netstat -s` in it's own key=value, using the line number as key and the line as value. The hack regarding the bold space is simply to hide the line numbers, and it is just exploiting some quirks in the way hardinfo works to do that.

lpereira:
Merged, thanks!

Issue #100, opened 2017-07-20
Column/Item label format conversion process


bp0:
(Starting in PR #98, this issue to track comments and progress) 

I think the labels of column headers and/or rows should be broken out of
the c-format strings because:
* The strings are more easily translated.
* The strings are more likely to be re-usable.
* A small change to one does not lose all of them.
* Changing a shell parameter does not lose all of them.
* Re-ordering them does not lose all of them.
* Adding items does not lose all of them.

I've been doing this in all new code, and converting old code.

Example:

	return g_strdup_printf(_("[$ShellParam$]\n"
			   "ReloadInterval=10000\n"
			   "ColumnTitle$TextValue=Name\n"
			   "ColumnTitle$Value=Group ID\n"
			   "ShowColumnHeaders=true\n"
			   "[Groups]\n"
			   "%s\n"), groups);

Changed to:

	return g_strdup_printf("[$ShellParam$]\n"
	               "ReloadInterval=10000\n"
	               "ColumnTitle$TextValue=%s\n" /* Name */
	               "ColumnTitle$Value=%s\n" /* Group ID */
	               "ShowColumnHeaders=true\n"
	               "[%s]\n%s\n",
	               _("Name"), _("Group ID"),
	               _("Groups"), groups);

Another example (post-conversion):

	gchar *val = g_strdup_printf("[%s]\n"
	        "%s=%d\n"
	        "%s=%d\n"
	        "%s=%s\n"
	        "%s=%s\n",
	        _("User Information"),
	        _("User ID"), (gint) passwd_->pw_uid,
	        _("Group ID"), (gint) passwd_->pw_gid,
	        _("Home Directory"), passwd_->pw_dir,
	        _("Default Shell"), passwd_->pw_shell);

# Status
- [x] modules/devices.c
- [x] modules/network.c
- [x] modules/computer.c
- [x] modules/benchmark.c
- [ ] modules/devices/spd-detect.c

PR #121, opened 2017-07-31
Tool to help save old label translations

bp0:
Used as part of https://github.com/lpereira/hardinfo/issues/100

After breaking the labels out of the c-format strings, this tool will help to save the old c-format string translations.

This is not an automated magical tool, it is an ugly hack of a helper for a tedious process. Read the instructions if you want to use it!


lpereira:
Merged, thanks!

Issue #105, opened 2017-07-20
Translation status


bp0:
This issue tracks the status of translating strings.
The list below is the output of `bash po/updatepo.sh` on master.
- [ ] de.po : (555 / 748 remain untranslated)
- [ ] es.po : (532 / 748 remain untranslated)
- [ ] fr.po : (512 / 748 remain untranslated)
- [ ] ru.po : (173 / 748 remain untranslated)

mckaygerhard:
do you think its necesary a manpage autocreation tool based by example in help2man? and then translated dinamically? on only let the english only manpage due the project its not so big to?

bp0:
I vote English only, simple to maintain. Does the command line interface even properly support translation? `LANGUAGE=ru hardinfo --help` is still in English for me.

mckaygerhard:
well we can generate a main englysh, later translate to.. umm wel as you note seems complicated now that --help output does not support translation!

mckaygerhard:
hi! i can see the about dialog has to many string depending of coding contribution.. can be made a parsing script tool that updates in every compilation taking that from AUHTORS.md file? i mean, due this lines of authors contributors are not translatables!
i just an idea due every commit that change any translated part, will let the xx.po files related outdated... and that's why i cannot concentrate in updated the languajes that i can contribute (English, some russian and greek)

bp0:
> i just an idea due every commit that change any translated part, will let the xx.po files related outdated...

I don't completely understand what you are saying, but if you are saying that old translations are lost when one translation changes, that really isn't true. Only that one string needs to be retranslated, and the old translation is added to the "obsolete strings" section of the .po file.
Whenever I change a translated string, I try to, and usually do, restore the old translated version.

> and that's why i cannot concentrate in updated the languajes that i can contribute (English, some russian and greek)

I would guess that English isn't your first language, but I would never have guessed it was Greek. :smile:  Don't you speak Spanish or Portuguese? I have the sense you are from South America.

mckaygerhard:
i have two problems.. the keyboards of the machines are too softly and i write too fast.. yes i 'm very lazy to write.. and i started to write in the typewriter machines.. so i hate the very soft way of the modern keyboards.. my mom ithe the "greek speaker" (well speak many languajes, i the only in my family that only speak two languajes), and ywe snif snif i'm from suth america... 

[...]

PR #122, opened 2017-08-01
Call setlocale() to clear the default "C" locale for CLI

bp0:
Fixes strings not being translated in the CLI.
See: https://github.com/lpereira/hardinfo/issues/105#issuecomment-319002090


lpereira:
Merged, thanks!

PR #125, opened 2017-08-06
GTK3 changes for syncmanager.c

bp0:
Missed it before because it is disabled by default. Plus, a couple other minor things.


lpereira:
Merged, thanks!

PR #126, opened 2017-08-06
computer.c: tweaks

bp0:
* Break "Not found" out of a c-format
* Tweak GDB regex. Old one would grab only closing paren


lpereira:
Merged, thanks!

PR #131, opened 2017-08-07
Making PCI devices information translatable

TotalCaesar659:

PR #130, opened 2017-08-07
Making locale information translatable

TotalCaesar659:

PR #129, opened 2017-08-07
Making module information translatable

TotalCaesar659:

PR #140, opened 2017-08-09
Expanded series based on TotalCaesar659's "Making ... translatable" PR series

bp0:
Continuation of @TotalCaesar659's #129, #130, #131, #132.
* more strings are translatable.
* labels are broken out of c-format strings as described at #100.
* some code cleanup and a few bug fixes along the way.

I did similar work on usb.c, but only tested the __scan_usb_lsusb() method. I can't test __scan_usb_procfs() or __scan_usb_sysfs() because those do not work on any machine I have access to.


lpereira:
Merged, thanks!

PR #141, opened 2017-08-10
Separate processor name and description + count cores and threads

bp0:
# CPU name de-prefixing and additional description
* add cpu_procs_cores_threads() function to get counts from sysfs/topology
* each platform must now provide processor_name() and processor_describe()
* processor_name_default(): returns a list of unique processor->model_name
* processor_describe_default(): returns "N physical; M cores; L threads"
* processor_describe_by_counting_names(): returns a list of unique
  processor->model_name with Nx prefix
  (ex: "4x ARM Cortex A53 + 4x ARM Cortex A33")
* x86: _name and _describe use defaults
* arm: _name returns name of SOC, if available, _describe returns
  processor_describe_by_counting_names()
* all other platforms: _name and _describe use defaults

# Clock frequncy description and benchmark result re-format
Current CPU configurations aren't properly represented in Hardinfo.
For SMT, each hardware thread is still reported as a CPU. Clusters
with different CPU clock rates are not reported. It is common for
ARM to pair a cluster of fast cores with a cluster of slower, but
more power-efficient cores. These changes attempt to address this.

The getProcessorFrequency method now returns the processor's max
frequency for all its cores. While the new
getProcessorFrequencyDesc lists each unique frequency with a prefix
Nx with the count of cores at that freqency.

Benchmark results have been reformated to use the de-prefixed
getProcessorName and getProcessorFrequencyDesc.

As an example from benchmark.conf:
`4x AMD Phenom(tm) II X4 940 Processor | 800 MHz`
becomes:
`AMD Phenom(tm) II X4 940 Processor | 4x 800 MHz`

Which, I think, makes much more sense, as it works well with
this kind of thing:

`Qualcomm Snapdragon 691 | 4x 1400 MHz + 4x 800 MHz`

# Screencaps
cpu_name_desc
soc_info
bench-result

# Issues
* "MHz" is not translated in benchmark.conf. This is an existing issue, but made more complicated now that the CPU Config string can't be atoi()-ed.


lpereira:
Merged, thanks!

Issue #142, opened 2017-08-11
Use JSON for benchmark.conf replacement


bp0:
More system information could be included and the results could be grouped by same machine. Something like simple [JSMN](http://zserge.com/jsmn.html) could be used to parse. Would probably make some future database sync easier as well.

After #141, could be something like:
```
{
  "board": < devices::getMotherboard >,
  "memory": < devices::getMemoryTotal >,
  "cpu_name": < devices::getProcessorName >,
  "cpu_desc": < devices::getProcessorDesc >,
  "cpu_config": < devices::getProcessorFrequencyDesc >,
  "results": {
         "CPU Zlib": 12.34,
         "FPU FFT": 78.91,
...
  }
}
```
Just an idea right now.

lpereira:
JSON could be used, yeah, but I'd rather use another JSON library to parse. JSMN isn't a parser, it's a lexical scanner; maybe something like json-glib would be better.

The format should be changed as well; if the server ends up being online again, the results for a particular benchmark will most likely come from different devices.

mckaygerhard:
its a great idea, inclusivelly usefully to inplement remote reports and communications.. 

but i think must finish some of the important task before starting with this, well i just my opinion..

bp0:
In #147, I just extended the current benchmark.conf format. There is new code that will make using a different format easier in the future, but for now I am satisfied and will not try and implement JSON. I like the JSMN better because the GObject-based solution is bulky and, I think, overkill for this application.

PR #147, opened 2017-08-14
Benchmark result details

bp0:
After considering JSON (#142) I decided to just extend the current format, and remain compatible with existing benchmark.conf formats.

* More information can be saved in benchmark.conf, while still being
  compatible with older versions.
* Selecting a result gives additional information about the result
  and the machine that produced it.

Showing an old result:
newbench_old
Showing a new result:
newbench_new


bp0:
@lpereira I have made the changes you suggest. 

The `mid` (machine id) field is generated from the data, it doesn't get decoded anytime later. It is used as the key in the key/value in the new benchmark.conf line format and for some ideas in the future.

Current quirks:
* Some translated strings are stored, potentially leading to strange mixes. CPU Description, "MHz" and "(Unknown)".
* It would be nice if the row to be selected by default could be specified somehow, so that the benchmark result is selected.

lpereira:
Merged, thanks!

PR #148, opened 2017-08-14
dmi: fail if only never succeeded in dmi_get_info_sys()

bp0:
Recent change to DMI causes it to fail when `/sys/class/dmi/id/product_family` is not available, which is not uncommon.


lpereira:
Merged, thanks!

PR #156, opened 2017-08-15
Remove duplicate group headers in computer summary

bp0:
As reported here: https://github.com/lpereira/hardinfo/issues/38#issuecomment-322467550


lpereira:
Merged, thanks!

Issue #154, opened 2017-08-15
Virtual Machine detection could be improved for VirtualBox


mckaygerhard:
i have several virtual box releases and virtual machine software/hardware for testing.. and some does not seem are detecting..

here a screen shot of virtualbox 3.2.4 russian that dont are detected in last git build, and the hardinfo report attached also:

![hardinfo-virtualboxdetection](https://user-images.githubusercontent.com/1571724/29318550-80686e9a-819e-11e7-89cc-8f9649883667.png)

[hardinfo_report.txt](https://github.com/lpereira/hardinfo/files/1225193/hardinfo_report.txt)

mckaygerhard:
attached corrected report, but still the previous its valid, the previous report was 2 months ago.. the attached here are today: bot detect the machine type wrong:

[hardinfo_report.txt](https://github.com/lpereira/hardinfo/files/1225243/hardinfo_report.txt)

bp0:
Can you try the tests from computer_get_virtualization() and see which test should succeed but doesn't?
https://github.com/lpereira/hardinfo/blob/master/modules/computer.c#L391

mckaygerhard:
well :

1. for virtual machines get info from /proc/cpuinfo does not make sense
2. seems in debian 8+ doe snot exits anymore /proc/scsi path but in my older etch there are !
3. quite strange, /var/log/dmesg has 0 bytes!!!!! 

i suggest let as is those files and added two more.. :

1. /proc/bus/input/devices: if at least one string are "irtual" (without Box) maybe could be a virtual machine, but if those are "mouse" device, its for sure are virtual machines.
2. if dmesg are not available can also read /var/log/messages as alternative..
3. lspci output! iknow its a stream and not a file, but if at least one string has "Inotek"/"ware" and also "irtu" its a virtual machine for sure!

bp0:
Does #157 solve this for you?

mckaygerhard:
sorry men, errr that pull only makes translatable the strings and added extra checks to model name.., the :
```
    tmp = module_call_method("devices::getMotherboard");
    if (strstr(tmp, "VirtualBox") != NULL) {
```
its the only related change and does not make any sense..  in fact i tested then and still does not work..

mckaygerhard:
i think all the ways to detect the machines must be based on varios string catures and not only in one/fisr provided/found

[...]

PR #157, opened 2017-08-15
Add another virtualbox test; make virt strings translatable

bp0:
Perhaps it will help with #154.

### DIFF
--- a/modules/computer.c
+++ b/modules/computer.c
@@ -403,35 +403,43 @@ gchar *computer_get_virtualization(void)
         gchar *vmtype;
     } vm_types[] = {
         /* VMware */
-        { "VMware", "Virtual (VMware)" },
-        { ": VMware Virtual IDE CDROM Drive", "Virtual (VMware)" },
+        { "VMware", N_("Virtual (VMware)") },
+        { ": VMware Virtual IDE CDROM Drive", N_("Virtual (VMware)") },
         /* QEMU */
-        { "QEMU", "Virtual (QEMU)" },
-        { "QEMU Virtual CPU", "Virtual (QEMU)" },
-        { ": QEMU HARDDISK", "Virtual (QEMU)" },
-        { ": QEMU CD-ROM", "Virtual (QEMU)" },
+        { "QEMU", N_("Virtual (QEMU)") },
+        { "QEMU Virtual CPU", N_("Virtual (QEMU)") },
+        { ": QEMU HARDDISK", N_("Virtual (QEMU)") },
+        { ": QEMU CD-ROM", N_("Virtual (QEMU)") },
         /* Generic Virtual Machine */
-        { ": Virtual HD,", "Virtual (Unknown)" },
-        { ": Virtual CD,", "Virtual (Unknown)" },
+        { ": Virtual HD,", N_("Virtual (Unknown)") },
+        { ": Virtual CD,", N_("Virtual (Unknown)") },
         /* Virtual Box */
-        { "VBOX", "Virtual (VirtualBox)" },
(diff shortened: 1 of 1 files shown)


modules/computer.c
```diff
@@ -403,35 +403,43 @@ gchar *computer_get_virtualization(void)
...
     DEBUG("Detecting virtual machine");
 
     if (g_file_test("/proc/xen", G_FILE_TEST_EXISTS)) {
          DEBUG("/proc/xen found; assuming Xen");
-         return g_strdup("Xen");
+         return g_strdup(_("Virtual (Xen)"));
+    }
+
+    tmp = module_call_method("devices::getMotherboard");
+    if (strstr(tmp, "VirtualBox") != NULL) {
```
mckaygerhard: umm i cannot remenber if this function are case-sensitive! but this will work!?

bp0:
Yes.

mckaygerhard:
i think this pull are not necesary, and sorry men, errr that pull only makes translatable the strings (more to translate) and added extra checks to model name.., the :
```
    tmp = module_call_method("devices::getMotherboard");
    if (strstr(tmp, "VirtualBox") != NULL) {
```
its the only related change and does not make any sense.. in fact i tested then and still does not work..

bp0:
It works because the DMI board name contains the string "VirtualBox" in VirtualBox.

mckaygerhard:
umm in fact yes! that's right as you can see in the pictures below, but still got unkwon phisical machine.. umm maybe i patch worng,, i'll download and compiles directly from your branch either got the patch

hardinfo-virtualbox

hardinfo-virtualbox-1

bp0:
Board/Name is actually missing in your screencap.

What does it show in English? `LANGUAGE=en hardinfo`
What is the actual board name value from DMI? `cat /sys/devices/virtual/dmi/id/board_name`
What version of VirtualBox is it?

mckaygerhard:
umm right!  as i said board_name its missing in this debian, or i mean, in the virtual machie interaction!?

but as i said are always present in the BIOS report name, so bios_vendor and produc_name are more viable way for sure .. only board name are not enought due in the Blade Vmware does not report enought info..  (i not have access right now but in one week i could test that):
```
root@sysinternet:/home/systemas/bin# ls /sys/devices/virtual/dmi/id/
bios_date     modalias	    product_serial   subsystem
bios_vendor   power	    product_uuid     sys_vendor
bios_version  product_name  product_version  uevent
root@sysinternet:/home/systemas/bin# 
systemas@sysinternet:~/bin$ cat /sys/devices/virtual/dmi/id/bios_vendor 
innotek GmbH
systemas@sysinternet:~/bin$ cat /sys/devices/virtual/dmi/id/product_name 
VirtualBox
systemas@sysinternet:~/bin$ 
```

bp0:
Can you try with combined #157 and #160 ?

mckaygerhard:
umm, its better to test with calm the #160 in few hours and i'll report feedback to you ... here i only have a russian machine with VB 3.2.4, inthe other section of the building have more to test.. gimme some time

mckaygerhard:
ok combined its working so relies on another function in another piece of code, the motherboard info as i suggested.. i true that duplication of code are not good but maybe if the code for DMI board manufacturer can be in same file?

.. also if the groups info already will provide headers so the headers section will have a unconsistent code.. to uinderstand

bp0:
The fix is actually that get_motherboard() will now fallback to dmi/product_name if dmi/board_name is empty. Your very old version of VirtualBox that only provided product_name is now detected.

bp0:
Rebased.

lpereira:
Merged, thanks!

PR #158, opened 2017-08-15
Update es.po to

mckaygerhard:


po/es.po
```diff
@@ -1663,19 +1663,19 @@ msgstr "Capacidades"
...
 
 #: modules/devices/arm/processor.c:339
 msgid "Clocks"
-msgstr ""
+msgstr "Ciclos"
 
 #: modules/devices/arm/processor.c:354
 msgid "SOC/Package Information"
-msgstr ""
+msgstr "Informacion SOC/Paquete"
```
lpereira: Shouldn't this have an accent? `Información`

mckaygerhard:
yeah, there's a missing accent here, i revised and seems some accent are missing..  but merged as is due i must updated again respect the new 8 pulls from bp0, or let it for later as you prefer

also a tip: when i revised the rusian hardinfo, i could see that XFce4 are very bad translated.. seems some strings are not same in context. 

so for what i mention: i can sure that my es.po are all in context and spell-related corrected!

po/es.po
mckaygerhard: yeah, there's a missing accent here, i revised and seems few only few accent are missing..  not so much..  also a tip: when i revised the rusian hardinfo, i could see that XFce4 are very bad translated.. seems some strings are not same in context.

all my pulls are editable by owers and collaborators, so if you or bp0 prefer can also edited and commited to ... 

so for what i mention: i can sure that my es.po are all in context and spell-related corrected!

mckaygerhard:
hi guys, some time passed, what about this spanish update laguaje pull?

lpereira:
Completely forgot about this one -- merged, ¡gracias!

PR #160, opened 2017-08-15
DMI: create single dmi_get_str() that uses sysfs with dmidecode fallback 

bp0:
This branch has really gone it's own course and ended up touching several things.
* dmi_util.h/c provides a interface to get dmi strings, dmi_get_str() and a function to get chassis type as a translated string, dmi_chassis_type_str(). It will try and use sysfs first, and if that fails, will use dmidecode, but that will usually only work as root.
* modeles/devices/dmi.c now only contains the report generating code, and uses dmi_util.
* dmi_util is linked to hardinfo and can be used by any module directly. I've also done this with dt_util and cpu_util, which contain similar functions that may be/are used across modules.
* devices and computer use dmi_util directly.
* the full set of dmidecode strings is now available.
* devices module: get_motherboard() has seen some improvements
* devices/dmi has more (all?) translatable strings.
* devices/dmi no longer suggests running as root if running as root


bp0:
Rebased.

hardinfo/dmi_util.c
```diff
@@ -0,0 +1,152 @@
...
+ *    You should have received a copy of the GNU General Public License
+ *    along with this program; if not, write to the Free Software
+ *    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hardinfo.h"
+#include "dmi_util.h"
+
+const char *dmi_sysfs_root() {
+    char *candidates[] = {
```
lpereira: These tables should always be `static const char *candidates[]`. However, for such a small table, it might be better to just write the function like so:

```C
static const char *dmi_sysfs_root(void)
{
        if (!access("/sys/devices/virtual/dmi", F_OK))
              return "/sys/devices/virtual/dmi";
        if (!access("/sys/class/dmi", F_OK))
              return "/sys/class/dmi";
        return NULL;
}
```

hardinfo/dmi_util.c
```diff
@@ -0,0 +1,152 @@
...
+    while (candidates[i] != NULL) {
+        if(access(candidates[i], F_OK) != -1)
+            return candidates[i];
+        i++;
+    }
+    return NULL;
+}
+
+char *dmi_get_str(const char *id_str) {
+  static struct {
```
lpereira: Weird indentation in this function: it should be 4 spaces.

hardinfo/dmi_util.c
```diff
@@ -0,0 +1,152 @@
...
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with this program; if not, write to the Free Software
+ *    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hardinfo.h"
+#include "dmi_util.h"
+
+const char *dmi_sysfs_root() {
```
lpereira: Functions that take no parameters should have `(void)`.  If they're only used within a file, they should be `static` as well, so they're not exposed.

hardinfo/dmi_util.c
```diff
@@ -0,0 +1,152 @@
...
+  gboolean spawned;
+  gchar *out, *err;
+
+  int i = 0;
+
+  /* try sysfs first */
+  if (dmi_root) {
+    while (tab_dmi_sysfs[i].id != NULL) {
+      if (strcmp(id_str, tab_dmi_sysfs[i].id) == 0) {
+        full_path = g_strdup_printf("%s/%s", dmi_root, tab_dmi_sysfs[i].path);
```
lpereira: Unless you need to store a path, always declare it as a array on the stack.  This reduces heap memory allocation, and it makes it easier to spot leaks (since it can't leak).  It can be `PATH_MAX` size and you can use `snprintf()` to populate it.

hardinfo/dmi_util.c
```diff
@@ -0,0 +1,152 @@
...
+    else
+      g_free(out);
+    g_free(err);
+  }
+
+dmi_str_done:
+  if (ret != NULL) {
+    ret = strend(ret, '\n');
+    ret = g_strstrip(ret);
+    if (strlen(ret) == 0) {
```
lpereira: No need to call `strlen()` if the only thing you want is to see if the string is empty: just use `if (!*ret)` instead.

hardinfo/dmi_util.c
```diff
@@ -0,0 +1,152 @@
...
+    if (chassis_type <= 0) {
+        gchar *chassis = dmi_get_str("chassis-type");
+        chassis_type = atoi(chassis);
+        g_free(chassis);
+    }
+
+    if (chassis_type >= 0 && chassis_type < G_N_ELEMENTS(types)) {
+        if (with_val)
+            return g_strdup_printf("[%d] %s", chassis_type, _(types[chassis_type]));
+        else
```
lpereira: Nitpicking, but `else` is not required if the previous `if` statement ended with a `return`.

hardinfo/dmi_util.c
```diff
@@ -0,0 +1,152 @@
...
+    if (strlen(ret) == 0) {
+      g_free(ret);
+      ret = NULL;
+    }
+  }
+  g_free(full_path);
+  return ret;
+}
+
+char *dmi_chassis_type_str(int chassis_type, int with_val) {
```
lpereira: `with_val` should be a `gboolean`.

includes/dmi_util.h
```diff
@@ -0,0 +1,12 @@
+#ifndef __DMI_UTIL_H__
```
lpereira: Please add the licensing boilerplate to this file.

modules/computer.c
```diff
@@ -303,38 +307,19 @@ static gchar *detect_machine_type(void)
...
-            N_("RAID Chassis"),
-            N_("Rack Mount Chassis"),
-            N_("Sealed-case PC"),
-        };
-        int chassis_type = atoi(idle_free(chassis));
-
-        if (chassis_type >= 0 && chassis_type < G_N_ELEMENTS(types))
-            return g_strdup(_(types[chassis_type]));
+    chassis = dmi_chassis_type_str(-1, 0);
+    if (chassis != NULL)
```
lpereira: `NULL` comparisons are often done explicitly in HardInfo: `if (chassis) ...`

modules/computer.c
```diff
@@ -303,38 +307,19 @@ static gchar *detect_machine_type(void)
...
+        return chassis;
+
+    chassis = dtr_get_string("/model", 0);
+    if (chassis) {
+        if (strstr(chassis, "Raspberry Pi") != NULL
+            || strstr(chassis, "ODROID") != NULL
+            /* etc */ ) {
```
lpereira: Instead of `/* etc */`, a comment such as `/* FIXME: consider making a table when adding more models */` would be nice

lpereira:
Merged, thanks!

Issue #162, opened 2017-08-16
Check usage of popen() for leaks via stderr


bp0:
As reported in issue #25, `popen()` can leak there is output to stderr and it hasn't been captured `(&2>1)`. It is probably better to use `g_spawn_command_line_sync()` and capture stdout and stderr separately.

- [ ] modules/network/samba.c:55:`    if ((usershare_list = popen("net usershare list", "r"))) {`
- [ ] modules/computer/boots.c:37:`    last = popen("last", "r");`
- [ ] modules/computer/languages.c:31:`    locale = popen("locale -va && echo", "r");`
- [ ] modules/computer/modules.c:53:`    lsmod = popen(lsmod_path, "r");`
- [ ] modules/computer/modules.c:79:`	modi = popen(buf, "r");`
- [ ] modules/network.c:93:`      if ((netstat = popen(command_line, "r"))) {`
- [ ] modules/network.c:191:`      if ((route = popen(command_line, "r"))) {`
- [ ] modules/network.c:267:`      if ((netstat = popen("netstat -an", "r"))) {`
- [ ] modules/devices/pci.c:58:`      if (!(lspci = popen(command_line, "r"))) {`
- [ ] modules/devices/pci.c:69:`      if (!(lspci = popen(tmp, "r"))) {`
- [x] modules/devices/dmi.c:83: (#160)
- [ ] modules/devices/battery.c:63:`    if (apcaccess_path && (apcaccess = popen(apcaccess_path, "r"))) {`
- [ ] modules/devices/storage.c:208:`		if ((prcap = popen(tmp, "r"))) {`
- [ ] modules/devices/usb.c:386:`    if (!(lsusb = popen(temp, "r"))) {`

lpereira:
Thinking about a macro that lets you write this:

```C
char *buffer;
FOR_EACH_STDOUT_LINE(buffer, "some-command %d", with_int_argument) {
    // Do something with buffer
}
```

This will:
- If there are more than 2 arguments, build the command line for `g_spawn_command_line_sync()` using `g_strdup_printf()` (and `idle_free()` the temporary), or pass the command line as is
- Check if the command was spawned properly, and if it was, loop using the same method mentioned in the original comment (`strchr()` looking for `\n`, etc.)
- Free memory allocated by `g_spawn_command_line_sync()` whenever it is done, possibly with `idle_free()` to make things easier

PR #164, opened 2017-08-16
samba.c: popen() leaks via stderr

bp0:
Part of #162: Change samba.c to use g_spawn_command_line_sync() instead of popen() to prevent leaks via stderr.
Prevents this message from appearing:
```
net usershare: cannot open usershare directory /var/lib/samba/usershares. Error No such file or directory
Please ask your system administrator to enable user sharing.
```
The problem is that I have no "usershares", so I didn't test if it is otherwise still working. Will someone who does test that?

### DIFF
--- a/modules/network/samba.c
+++ b/modules/network/samba.c
@@ -31,7 +31,7 @@ scan_samba(void)
 {
     gchar *str;
     gsize length;
-    
+
     if (smb_shares_list) {
         g_free(smb_shares_list);
         smb_shares_list = g_strdup("");
@@ -41,7 +41,7 @@ scan_samba(void)
                             &str, &length, NULL)) {
         shell_status_update("Scanning SAMBA shares...");
         scan_samba_from_string(str, length);
-        g_free(str);                        
+        g_free(str);
     }
 
     scan_samba_usershares();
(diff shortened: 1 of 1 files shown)


modules/network/samba.c
```diff
@@ -51,33 +51,36 @@ void
...
-        
+    gchar *out, *err;
+    gboolean spawned;
+    int status;
+    gchar *p, *next_nl;
+    gchar *usershare, *cmdline;
+    gsize length;
+
+    spawned = g_spawn_command_line_sync("net usershare list",
+            &out, &err, &status, NULL);
```
lpereira: If `stderr` is not required, you can pass `NULL` instead of `&err`; then there's no need to `free()` it below.

bp0: It appears that if you pass NULL, then stderr will still leak.

lpereira: If it's leaking, it's a bug in GLib. It's marked as [optional](https://developer.gnome.org/glib/stable/glib-Spawning-Processes.html#g-spawn-command-line-sync) in the documentation.

bp0: If NULL is passed, the following message appears in Terminal when Shared Directories is selected:
```
net usershare: cannot open usershare directory /var/lib/samba/usershares. Error No such file or directory
Please ask your system administrator to enable user sharing.
```
I don't know, but I think it could still be optional and just forward stderr to the application's stderr if it is NULL. 
When it is captured in a string, the message does not appear in Terminal.

lpereira: Ah, I see. LGTM, then.

lpereira:
Merged, thanks!

PR #165, opened 2017-08-16
boots.c: capture stderr to prevent a leak

bp0:
Part of #162.
boots.c uses `last` to get the list of recent boots. Redirect stderr for popen() to prevent leak via stderr.
The only error would be inability to access /var/log/wtmp.

Also, removes double group header now that info_*() will provide one.

### DIFF
--- a/modules/computer/boots.c
+++ b/modules/computer/boots.c
@@ -24,43 +24,43 @@
 void
 scan_boots_real(void)
 {
-    FILE *last;
-    char buffer[256];
+    gchar **tmp;
+    gboolean spawned;
+    gchar *out, *err, *p, *s, *next_nl;
 
     scan_os(FALSE);
 
     if (!computer->os->boots)
-      computer->os->boots = g_strdup_printf("[%s]\n", _("Boots"));
+      computer->os->boots = strdup("");
     else
       return;
 
-    last = popen("last", "r");
-    if (last) {
-      while (fgets(buffer, 256, last)) {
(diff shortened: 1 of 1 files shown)


modules/computer/boots.c
```diff
@@ -30,11 +30,11 @@ scan_boots_real(void)
     scan_os(FALSE);
 
     if (!computer->os->boots)
-      computer->os->boots = g_strdup_printf("[%s]\n", _("Boots"));
+      computer->os->boots = g_strdup_printf("");
     else
       return;
 
-    last = popen("last", "r");
+    last = popen("last 2>&1", "r");
```
lpereira: I'd rather just use `g_spawn_sync_with_pipes()` everywhere instead of `popen()`, which spawns a shell. The glib function will `fork()` and `exec()` by itself.

modules/computer/boots.c
```diff
@@ -30,11 +30,11 @@ scan_boots_real(void)
     scan_os(FALSE);
 
     if (!computer->os->boots)
-      computer->os->boots = g_strdup_printf("[%s]\n", _("Boots"));
+      computer->os->boots = g_strdup_printf("");
```
lpereira: It's OK to leave this like so; it looks weird, but it's a reminder to fully convert this to use `struct Info`.

bp0: That should be just strdup() anyway.

lpereira:
Merged, thanks!

PR #167, opened 2017-08-16
Remove the appended "Module" from about module dialog title

bp0:
The string is constructed one word at a time, "About", X, "Module"
and it ends up being very strange in languages other than English.
Cutting out the module makes it a little better, and not much is
lost. A module could add "Module" to it's title, and then at
least that string can be translated as a complete unit. 
See https://github.com/lpereira/hardinfo/issues/38#issuecomment-322625707


lpereira:
Merged, thanks!

PR #169, opened 2017-08-17
languages.c: popen() -> g_spawn_command_line_sync(), capture stderr

bp0:
Part of #162.
I don't know if this would ever leak.


bp0:
Requested changes applied.

PR #170, opened 2017-08-17
fix detection of OpenGL profiles and ARB_compatibility

### LINKED ISSUE #27: doesn't properly identify opengl version
My hardinfo reports incorrect version of opengl

mckaygerhard:
OpenGL version 3.0 introduced the idea of deprecation (ah rare, umm again another deprecation ifff) so they can be remove some features on olders specification in later versions, which happened in version 3.1, when the Fixed Function Pipeline was taken out.

However, since many implementations support the deprecated and removed features anyway, and it has settled down into a division between Core and Compatibility

today the "core" are the real OpenGL version and the "other older" are the compatibility version.. 

the pull fixed #27 and closes #27 due are able to work for older machines and newer.. i'll explain: 

its suppost that hardinfo its a soft&hard detection tool, if i put my livecd with hardinfo into my very great powered Dell XPS but with my own compiled Xorg very older (yeah steam consumes too many resources so use a older version has more performance).. if that patch dont take in consideration older and newer Mesa ARB_compatibility/Profiles hardinfo cannot detect the "mixed OpenGL environment" version..
this behaviour are very common in linux such as Alpine, Gentoo, Funto (my favorite of this king) or all of then that are compile at machine.. and not binary-distribute

well for what i need a old xorg with newer linux, ok try to play a guindows newer game ported into steam.. its very slow respect guindo release.. some gamers like me combines the packages and compilations to gain the power of performance..


modules/computer/display.c
```diff
@@ -28,18 +28,22 @@ get_glx_info(DisplayInfo *di)
...
+	    {
+		if ( strstr(output_lines[i], "profile") ) {    cpd = 1;    }
 		gchar **tmp = g_strsplit(output_lines[i], ":", 0);
 
 		tmp[1] = g_strchug(tmp[1]);
 
 		get_str("OpenGL vendor str", di->ogl_vendor);
 		get_str("OpenGL renderer str", di->ogl_renderer);
-		get_str("OpenGL version str", di->ogl_version);
+		if ( corep > 0 ) {    get_str("OpenGL core profile version str", di->ogl_version);    }
```
bp0: corep or cpd?

bp0:
I think it may be better to just capture and report all version strings.

modules/computer/display.c
mckaygerhard: cpd cpd! ok ok its due i was triying in varous machines, you can made modifications on my pull directly.. github have the feature that "able to made modification by others"

mckaygerhard:
yeah! i have the code but the changes are over others and some headers.. due must show more lines the structure definition will change.. 

so then for no mayor modification firs, fix the issue and then made the feature, i mean one step at one

modules/computer/display.c
```diff
@@ -28,18 +28,22 @@ get_glx_info(DisplayInfo *di)
...
 	gint i = 0;
+	gint cpd = 0;
 
 	for (output_lines = g_strsplit(output, "\n", 0);
 	     output_lines && output_lines[i];
 	     i++) {
-	    if (strstr(output_lines[i], "OpenGL")) {
+	    if (strstr(output_lines[i], "OpenGL")) 
+	    {
+		if ( strstr(output_lines[i], "profile") ) {    cpd = 1;    }
```
lpereira: I know HardInfo is a mess regarding the coding style, but newer code should be more consistent. Please ensure that:

- There's no space after `(` or before `)`
- `if` statements with a single statement do not get braces
- `if` statements should be on a line by themselves

modules/computer/display.c
```diff
@@ -28,18 +28,22 @@ get_glx_info(DisplayInfo *di)
     if (g_spawn_command_line_sync("glxinfo", &output, NULL, NULL, NULL)) {
 	gchar **output_lines;
 	gint i = 0;
+	gint cpd = 0;
```
lpereira: This could be a `gboolean`.  Please also name it `has_core_profile` or something of the sort. Shorter variable names should only be used for things that are commonly used (such as loop induction variables, `i`, `j`, etc).

mckaygerhard:
> In addition to code style issues, is there a fundamental difference between the OpenGL core profile version and the OpenGL version? Should HardInfo care?

seems that the "core profile" will be the default in some years.. of course, i only use the old string due all my xorg versions are below 1.7 and mesa below 3.0 so in my cases i dont have "core profile".. 

now i going to make your requested changes..

modules/computer/display.c
mckaygerhard: fixed

modules/computer/display.c
mckaygerhard: fixed, i must point that the code was made before the source styling fixation..that's why ...

lpereira:
Please rebase and push again.  Also, please notice the coding style in the `if` statements:

```C
               if ( corep > 0 ) {    get_str("OpenGL core profile version str", di->ogl_version);    }
                 else             {    get_str("OpenGL version str", di->ogl_version);    }
```

Should be written as:

```C
             if (corep > 0)
                    get_str(...);
             else
                    get_str(...);
```

Notice the line breaks, indentation, and spaces.

mckaygerhard:
the code of conditional around if-else was fixed some time ago, and now i rebased

mckaygerhard:
hi guys, some time passed, i rebase some time ago but this patch till applied

mckaygerhard:
@bp0  and @lpereira , important news: i made some tests in middle age hardware and in fresh lasted hardware,  respect the:
> In addition to code style issues, is there a fundamental difference between the OpenGL core profile version and the OpenGL version? Should HardInfo care?

the new core profile its due OpenGL 3.0 introduced the concept of deprecation. A whole bunch of functionallity that can be implemented in software on either the CPU side or in shaders got marked for removal in future versions (i.e. 3.1).

**The word compatibillity is meant in regard to older software, not older hardware. Special case, CAD vendors didn't want to rewrite their decade old cruft as example, or very popular game engines such those derivates from quake and unreal...  this change are due stupid windoze hardware due now OpenGL are very popular in guiindows**


### About the OPenGL

Core profile means OpenGL(R) minus the deprecated functions, compatibillity profile means OpenGL(R) with the deprecated functions NOT deprecated hardware

### about the OpenCL

the **opencl are interfaces for manage and permit to use GPU's as CPU's** in easy words! i was thinki8ng some time ago that will replace opengl in future but not, i was wrong, that was started by AMD/ATI some time ago https://es.wikipedia.org/wiki/Close_to_Metal

### how was?, why? opengl core profile are more newer: NO!

those was due, i found, i see a xtrage versioning in my recent hardware, respect the #27 that shows OpenGL more older, in my case was more recent, as can see:

``` bash
glxinfo | grep version
server glx version string: 1.4
client glx version string: 1.4
GLX version: 1.4
OpenGL version string: 2.1 Mesa 13.0.6
OpenGL shading language version string: 1.20
OpenGL ES profile version string: OpenGL ES 2.0 Mesa 13.0.6
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 1.0.16
```

bp0:
So after OpenGL 3.1, "OpenGL version string" became the compatibility profile string, and "OpenGL core profile version string" became the new version string. So the compatibility profile string will never be more than 3.0 (and similarly GLSL 1.3). 
There are several things that need doing with this, for example there are now commonly multiple cards from different vendors and things like GLVND.

bp0:
This commit could be accepted, but a better solution is needed. See #225.

mckaygerhard:
@bp0  of course but the case its that BOTH information versions must be showed? not at all, the problem its that most users think that "core profile" its the only future and was not true.. core profile its only a step area to move some deprecations.. i was explained that..

mckaygerhard:
the compatiblity contex due many of the packages installed in OS need it!, if not that version "core" profile will not happend, 

_BOTH information versions must be showed? not at all_, the problem its that most users think that "core profile" its the only future and was not true.. core profile its only a step area to move some deprecations.. i was explained that.. **as my mesa its more recent now but core profile are not showed**:

``` bash
glxinfo | grep version
server glx version string: 1.4
client glx version string: 1.4
GLX version: 1.4
OpenGL version string: 2.1 Mesa 13.0.6
OpenGL shading language version string: 1.20
OpenGL ES profile version string: OpenGL ES 2.0 Mesa 13.0.6
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 1.0.16
```

bp0:
This patch is not applicable after ea8ce7dabdf2e3a27e1d6edb84752dff5700a849. 
See `fill_glx_info()` in `harinfo/x_util.c` and `callback_display()` in `modules/computer.c`.
If you think the GLX information is still reported incorrectly, please make a new issue or pull request.

mckaygerhard:
ok the new Graphics feature module has proper identification? if yes then i'll test and report feedback to the issue #27

PR #168, opened 2017-08-16
Fix for compilation warning

TotalCaesar659:
Here is a fix for the compilation warning:
```
/home/max/hardinfo/modules/benchmark.c: In function ‘do_benchmark’:
/home/max/hardinfo/modules/benchmark.c:400:8: warning: ‘gtk_message_dialog_set_image’ is deprecated [-Wdeprecated-declarations]
        gtk_message_dialog_set_image(GTK_MESSAGE_DIALOG(bench_dialog), bench_image);
        ^
In file included from /usr/include/gtk-3.0/gtk/gtk.h:143:0,
                 from /home/max/hardinfo/includes/hardinfo.h:22,
                 from /home/max/hardinfo/modules/benchmark.c:19:
/usr/include/gtk-3.0/gtk/gtkmessagedialog.h:115:12: note: declared here
 void       gtk_message_dialog_set_image    (GtkMessageDialog *dialog,
            ^
[ 65%] Building C object CMakeFiles/benchmark.dir/modules/benchmark/blowfish.c.o
```
I didn't see any difference in GTK2 and GTK3 versions after and before this PR.

PR #172, opened 2017-08-17
Re-fix for compilation warning

TotalCaesar659:
Re-fix for compilation warning https://github.com/lpereira/hardinfo/pull/168.

### DIFF
--- a/modules/benchmark.c
+++ b/modules/benchmark.c
@@ -388,6 +388,27 @@ static void do_benchmark(void (*benchmark_function)(void), int entry)
        bench_image = icon_cache_get_image("benchmark.png");
        gtk_widget_show(bench_image);
 
+#if GTK_CHECK_VERSION(3, 0, 0)
+       GtkWidget *button;
+       GtkWidget *content_area;
+       GtkWidget *hbox;
+       GtkWidget *label;
+       
+       bench_dialog = gtk_dialog_new_with_buttons("",
+                                                  NULL,
+                                                  GTK_DIALOG_MODAL,
+                                                  _("Cancel"),
+                                                  GTK_RESPONSE_ACCEPT,
+                                                  NULL);
+       content_area = gtk_dialog_get_content_area(GTK_DIALOG(bench_dialog));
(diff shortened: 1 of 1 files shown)


modules/benchmark.c
```diff
@@ -388,6 +388,27 @@ static void do_benchmark(void (*benchmark_function)(void), int entry)
...
        gtk_widget_show(bench_image);
 
+#if GTK_CHECK_VERSION(3, 0, 0)
+       GtkWidget *button;
+       GtkWidget *content_area;
+       GtkWidget *hbox;
+       GtkWidget *label;
+       
+       bench_dialog = gtk_dialog_new_with_buttons("",
+                                                  NULL,
```
lpereira: Shouldn't this second parameter be the parent window (`GTK_WINDOW(shell_get_main_shell()->window)`)? Otherwise, this dialog won't be modal.

If the GTK+3 code works in GTK+2, we can just get rid of both versions and keep the one that works in GTK+3.

TotalCaesar659: Well, GTK3 code doesn't work with GTK2 build, it has new warning when compiling:
```
/home/max/hardinfo/modules/benchmark.c: In function ‘do_benchmark’:
/home/max/hardinfo/modules/benchmark.c:438:13: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
        hbox = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, 5);
             ^
```
And HardInfo crashes after choosing a benchmark. I think `gtk_box_new` doesn't work with GTK2, so it should be better to merge it "as is" with changing only `NULL` to `GTK_WINDOW(shell_get_main_shell()->window)` as you advised.

TotalCaesar659: @lpereira Please merge this PR with replacing `NULL` to `GTK_WINDOW(shell_get_main_shell()->window)`, it works both GTK2 and 3. It should be the best way, I guess.

lpereira:
Merged, thanks!

PR #173, opened 2017-08-17
Making button translatable

TotalCaesar659:


bp0:
The button should still be translatable with the "_" in there. That just sets the keyboard shortcut key, Alt+C for "_Cancel". You don't need to have an "_" in the translated string, but you could, and whatever character came after would be the shortcut key.

TotalCaesar659:
Yeah, fixed on https://github.com/lpereira/hardinfo/pull/176.

PR #175, opened 2017-08-18
usb_util.c

bp0:
A set of functions for getting information for a single USB device,
or a list of all devices.

The only implemented method is using `lsusb`, which is slow. A
method using sysfs would be much better. The existing sysfs and
procfs methods in devices/usb.c do not appear to work, so it would
have to be something new.

devices/usb.c modified to use usb_util, but all the old code is
still there.

Solves #162 for devices/usb.c
Before it was common for `lsusb` to leak the message below once for each device.
`Couldn't open device, some information will be missing`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -215,6 +215,7 @@ endif()
 
 if (HARDINFO_GTK3)
 add_executable(hardinfo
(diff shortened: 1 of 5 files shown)


bp0:
This is too slow right now, closing.

PR #178, opened 2017-08-19
usb_util.c [v2]

bp0:
A set of functions for getting information for a single USB device,
or a list of all devices.

The only implemented method is using `lsusb`, which is slow. A
method using sysfs would be much better. The existing sysfs and
procfs methods in devices/usb.c do not appear to work, so it would
have to be something new.

devices/usb.c modified to use usb_util, but all the old code is
still there.

Solves #162 for devices/usb.c
Before it was common for `lsusb` to leak the message below once for each device.
`Couldn't open device, some information will be missing`

Fixed a problem with #175.

### DIFF
--- a/hardinfo/usb_util.c
+++ b/hardinfo/usb_util.c
@@ -22,22 +22,18 @@
 #include "usb_util.h"
 
 usbd *usbd_new() {
-    usbd *s = malloc(sizeof(usbd));
-    if (s) {
-        memset(s, 0, sizeof(usbd));
-    }
-    return s;
+    return g_new0(usbd, 1);
 }
 
 void usbd_free(usbd *s) {
     if (s) {
-        free(s->vendor);
-        free(s->product);
-        free(s->usb_version);
-        free(s->device_version);
-        free(s->dev_class_str);
-        free(s->dev_subclass_str);
-        free(s);
+        g_free(s->vendor);
+        g_free(s->product);
+        g_free(s->usb_version);
+        g_free(s->device_version);
+        g_free(s->dev_class_str);
+        g_free(s->dev_subclass_str);
+        g_free(s);
     }
 }
 
@@ -51,7 +47,7 @@ void usbd_list_free(usbd *s) {
 }
 
 /* returns number of items after append */
-int usbd_list_append(usbd *l, usbd *n) {
+static int usbd_list_append(usbd *l, usbd *n) {
     int c = 0;
     while(l != NULL) {
         c++;
@@ -71,15 +67,15 @@ int usbd_list_count(usbd *s) {
     return usbd_list_append(s, NULL);
 }
 
-char *_lsusb_lv(char *line, const char *prefix) {
-    if ( g_str_has_prefix(line, prefix) ) {
+static char *lsusb_line_value(char *line, const char *prefix) {
+    if (g_str_has_prefix(line, prefix)) {
         line += strlen(prefix) + 1;
         return g_strstrip(line);
     } else
         return NULL;
 }
 
-int usb_get_device_lsusb(int bus, int dev, usbd *s) {
+static gboolean usb_get_device_lsusb(int bus, int dev, usbd *s) {
     gboolean spawned;
     gchar *out, *err, *p, *l, *t, *next_nl;
     gchar *lsusb_cmd = g_strdup_printf("lsusb -s %d:%d -v", bus, dev);
@@ -92,34 +88,34 @@ int usb_get_device_lsusb(int bus, int dev, usbd *s) {
     g_free(lsusb_cmd);
     if (spawned) {
         if (strstr(err, "information will be missing")) {
-            s->user_scan = 1;
+            s->user_scan = TRUE;
         }
         p = out;
         while(next_nl = strchr(p, '\n')) {
             strend(p, '\n');
             g_strstrip(p);
-            if (l = _lsusb_lv(p, "idVendor")) {
+            if (l = lsusb_line_value(p, "idVendor")) {
                 s->vendor_id = strtol(l, NULL, 0);
                 if (t = strchr(l, ' '))
-                    s->vendor = strdup(t + 1);
-            } else if (l = _lsusb_lv(p, "idProduct")) {
+                    s->vendor = g_strdup(t + 1);
+            } else if (l = lsusb_line_value(p, "idProduct")) {
                 s->product_id = strtol(l, NULL, 0);
                 if (t = strchr(l, ' '))
-                    s->product = strdup(t + 1);
-            } else if (l = _lsusb_lv(p, "MaxPower")) {
+                    s->product = g_strdup(t + 1);
+            } else if (l = lsusb_line_value(p, "MaxPower")) {
                 s->max_curr_ma = atoi(l);
-            } else if (l = _lsusb_lv(p, "bcdUSB")) {
-                s->usb_version = strdup(l);
-            } else if (l = _lsusb_lv(p, "bcdDevice")) {
-                s->device_version = strdup(l);
-            } else if (l = _lsusb_lv(p, "bDeviceClass")) {
+            } else if (l = lsusb_line_value(p, "bcdUSB")) {
+                s->usb_version = g_strdup(l);
+            } else if (l = lsusb_line_value(p, "bcdDevice")) {
+                s->device_version = g_strdup(l);
+            } else if (l = lsusb_line_value(p, "bDeviceClass")) {
                 s->dev_class = atoi(l);
                 if (t = strchr(l, ' '))
-                    s->dev_class_str = strdup(t + 1);
-            } else if (l = _lsusb_lv(p, "bDeviceSubClass")) {
+                    s->dev_class_str = g_strdup(t + 1);
+            } else if (l = lsusb_line_value(p, "bDeviceSubClass")) {
                 s->dev_subclass = atoi(l);
                 if (t = strchr(l, ' '))
-                    s->dev_subclass_str = strdup(t + 1);
+                    s->dev_subclass_str = g_strdup(t + 1);
             }
             /* TODO: speed_mbs
              * WISHLIST: interfaces, drivers */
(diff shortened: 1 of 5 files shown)


bp0:
Rebased.

hardinfo/usb_util.c
```diff
@@ -0,0 +1,185 @@
...
+        l = l->next;
+    }
+    return c;
+}
+
+int usbd_list_count(usbd *s) {
+    return usbd_list_append(s, NULL);
+}
+
+char *_lsusb_lv(char *line, const char *prefix) {
```
lpereira: Internal functions should always be marked as `static`. There's no need to prepend an underscore in their names. I'd prefer if their names were more descriptive as well: I have no idea what `lv` means here.

hardinfo/usb_util.c
```diff
@@ -0,0 +1,185 @@
...
+    if ( g_str_has_prefix(line, prefix) ) {
```
lpereira: No spaces after or before parenthesis: `if (g_str_has_prefix(...)) {`.

hardinfo/usb_util.c
```diff
@@ -0,0 +1,185 @@
...
+        line += strlen(prefix) + 1;
+        return g_strstrip(line);
+    } else
+        return NULL;
+}
+
+int usb_get_device_lsusb(int bus, int dev, usbd *s) {
+    gboolean spawned;
+    gchar *out, *err, *p, *l, *t, *next_nl;
+    gchar *lsusb_cmd = g_strdup_printf("lsusb -s %d:%d -v", bus, dev);
```
lpereira: I wonder if we would benefit from a `h_popen_printf()` that worked pretty much like a `printf()` function. Maybe with `fopencookie()`, this could be used like `popen()`:

```C
    FILE *lsusb = h_popen_printf("lsusb -s %d:%d -v", bus, dev);
    char buffer[1024];

    if (!lsusb)
        return -errno;

    while (fgets(buffer, sizeof(buffer), lsusb)) {
         ...
    }

    h_popen_close(lsusb);
```

This would simplify a lot all these places where `strchr()` and `strend()` are used inside a loop: these could be moved to one of the cookie functions.

(No need to do this for this patch, just thinking out loud.)

hardinfo/usb_util.c
```diff
@@ -0,0 +1,185 @@
...
+                if (t = strchr(l, ' '))
+                    s->dev_subclass_str = strdup(t + 1);
+            }
+            /* TODO: speed_mbs
+             * WISHLIST: interfaces, drivers */
+            p = next_nl + 1;
+        }
+        g_free(out);
+        g_free(err);
+        return 1;
```
lpereira: If a function is returning only 1 or 0, it should return a boolean value instead.

modules/devices/usb.c
```diff
@@ -425,11 +426,92 @@ gboolean __scan_usb_lsusb(void)
...
 }
 
-void __scan_usb(void)
-{
-    if (!__scan_usb_procfs()) {
-        if (!__scan_usb_sysfs()) {
-             __scan_usb_lsusb();
+#define UNKIFNULL_AC(f) (f != NULL) ? f : _("(Unknown)");
+
+void _usb_dev(const usbd *u) {
```
lpereira: `static`

modules/devices/usb.c
```diff
@@ -425,11 +426,92 @@ gboolean __scan_usb_lsusb(void)
...
+                );
+
+    moreinfo_add_with_prefix("DEV", key, str); /* str now owned by morinfo */
+
+    g_free(v_str);
+    g_free(name);
+    g_free(key);
+}
+
+gboolean __scan_usb_util(void) {
```
lpereira: `static`

modules/devices/usb.c
```diff
@@ -425,11 +426,92 @@ gboolean __scan_usb_lsusb(void)
...
+
+    if (c > 0) {
+        if (usb_list) {
+            moreinfo_del_with_prefix("DEV:USB");
+            g_free(usb_list);
+        }
+        usb_list = g_strdup_printf("[%s]\n", _("USB Devices"));
+
+        while(curr) {
+            //printf("USB: Bus %03d Dev %03d Ven %04x Prod %04x %s %s\n",
```
lpereira: Please remove debugging comments.

includes/usb_util.h
```diff
@@ -0,0 +1,35 @@
+#ifndef __USB_UTIL_H__
```
lpereira: Missing license boilerplate.

includes/usb_util.h
```diff
@@ -0,0 +1,35 @@
...
+#define __USB_UTIL_H__
+
+/* this is a linked list */
```
lpereira: Why hand-roll a linked list and not use a GSList?

includes/usb_util.h
```diff
@@ -0,0 +1,35 @@
...
+    char *dev_subclass_str;
+
+    char *usb_version;
+    char *device_version; /* bcdDevice */
+
+    int max_curr_ma;
+
+    int speed_mbs; /* TODO: */
+
+    int user_scan; /* not scanned as root */
```
lpereira: Why not a boolean?

hardinfo/usb_util.c
```diff
@@ -0,0 +1,185 @@
...
+    usbd *head = NULL, *nd;
+    int bus, dev, vend, prod, ec;
+
+    spawned = g_spawn_command_line_sync("lsusb",
+            &out, &err, NULL, NULL);
+    if (spawned) {
+        p = out;
+        while(next_nl = strchr(p, '\n')) {
+            strend(p, '\n');
+            if (ec = sscanf(p, "Bus %d Device %d: ID %x:%x", &bus, &dev, &vend, &prod) ) {
```
lpereira: Won't this issue a warning? You're likely to need two parenthesis here. Or, better yet, just write in two lines:

```C
ec = sscanf(...);
if (ec == 4) {
    ...
}
```

hardinfo/usb_util.c
```diff
@@ -0,0 +1,185 @@
...
+    usbd *n;
+    while(s != NULL) {
+        n = s->next;
+        usbd_free(s);
+        s = n;
+    }
+}
+
+/* returns number of items after append */
+int usbd_list_append(usbd *l, usbd *n) {
```
lpereira: Why not just use a `GSList`?

hardinfo/usb_util.c
```diff
@@ -0,0 +1,185 @@
...
+ *    You should have received a copy of the GNU General Public License
+ *    along with this program; if not, write to the Free Software
+ *    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hardinfo.h"
+#include "usb_util.h"
+
+usbd *usbd_new() {
+    usbd *s = malloc(sizeof(usbd));
```
lpereira: Use `g_new0()` instead. It already checks if the allocation fails and will most likely allocate memory from a pool that has been `memset()` to zero already (usually fresh from the kernel).  Be sure to use `g_free()` for things allocated with `g_new()`.

hardinfo/usb_util.c
```diff
@@ -0,0 +1,185 @@
...
+        if (strstr(err, "information will be missing")) {
+            s->user_scan = 1;
+        }
+        p = out;
+        while(next_nl = strchr(p, '\n')) {
+            strend(p, '\n');
+            g_strstrip(p);
+            if (l = _lsusb_lv(p, "idVendor")) {
+                s->vendor_id = strtol(l, NULL, 0);
+                if (t = strchr(l, ' '))
```
lpereira: This will generate a warning. You need two parenthesis enclosing the expression here: `if ((t = strchr(...)))`

hardinfo/usb_util.c
```diff
@@ -0,0 +1,185 @@
...
+                    s->vendor = strdup(t + 1);
```
lpereira: Please use `g_strdup()`, it already NULL-checks the return value and aborts the program if allocation fails

hardinfo/usb_util.c
```diff
@@ -0,0 +1,182 @@
...
+                c++;
+            }
+            break;
+        }
+        l = l->next;
+    }
+    return c;
+}
+
+int usbd_list_count(usbd *s) {
```
lpereira: It's very surprising that a function to count the elements of a linked list actually modifies it.  This should really be just a wrapper around GSList.

includes/usb_util.h
```diff
@@ -0,0 +1,53 @@
...
+ *    You should have received a copy of the GNU General Public License
+ *    along with this program; if not, write to the Free Software
+ *    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#ifndef __USB_UTIL_H__
+#define __USB_UTIL_H__
+
+/* this is a linked list */
+typedef struct usbd {
```
lpereira: For things that are global identifiers, it's better to have "namespaces", or at least longer names.  Shorter names are useful for static functions or local symbols within functions.

mckaygerhard:
hi now that some review was made .. this will be compatible with some older hardware? i mean this uses sysfs, what happened with those older systems that are kernel usage << 3.0.24 ? for some embebed linux

hardinfo/usb_util.c
bp0: Because usbd_list_append() appends an element and returns the length, usbd_list_count() can just call it without a new element and get the length. append is already stepping through the list, so I just re-used it.

includes/usb_util.h
bp0: Would something like usb_device_info be enough?

bp0:
@mckaygerhard, this adds a new method, but falls back to the previously existing methods. If they ever worked, which I have not been able to confirm, then they should still work.

lpereira:
Merged, thanks!

PR #179, opened 2017-08-19
info_add_group_ff(), info.c-related ideas

bp0:
This is an idea to make it easier to convert the existing sprintf()s to using the Info structure. It's just a version of info_add_group() that takes name/value pairs instead of InfoField structures.
All the value strings are duplicated, and freed, so something like idle_free() has to be used for temp strings, but I was thinking maybe we can work that into a formatting function that would serve like info_field_printf().
Would look something like this:
```
info_add_group_ff(info, 
     _("Group"),
     _("Easy One"), some->value_freed_elsewhere, 
     _("Clock"), _fv("%d %s", freq, _("MHz") ), 
     _("Something"), _fv(tmp_value), 
```
And _fv() is the equivalent of:
```
    idle_free(g_strdup_printf(fmt, ...))
```

Part of #180.

### DIFF
--- a/hardinfo/info.c
+++ b/hardinfo/info.c
@@ -56,6 +56,33 @@ void info_add_group(struct Info *info, const gchar *group_name, ...)
     g_array_append_val(info->groups, group);
 }
 
+void info_add_group_ff(struct Info *info, const gchar *group_name, ...)
+{
+    struct InfoGroup group = {
+        .name = group_name,
+        .fields = g_array_new(FALSE, FALSE, sizeof(struct InfoField))
+    };
+    va_list ap;
+
+    va_start(ap, group_name);
+    while (1) {
+        struct InfoField field = {
+            .name = va_arg(ap, char *),
+            .value = va_arg(ap, char *),
+            .free_value_on_flatten = TRUE,
+        };
+        if (field.value)
+            field.value = g_strdup(field.value);
+
+        if (!field.name)
+            break;
+        g_array_append_val(group.fields, field);
+    }
+    va_end(ap);
+
+    g_array_append_val(info->groups, group);
+}
+
 struct InfoField info_field(const gchar *name, gchar *value)
 {
     return (struct InfoField) {
--- a/hardinfo/info.c
+++ b/hardinfo/info.c
@@ -181,10 +181,17 @@ static void flatten_group(GString *output, const struct InfoGroup *group)
     if (group->fields) {
         for (i = 0; i < group->fields->len; i++) {
             struct InfoField field;
+            gchar *tag;
 
             field = g_array_index(group->fields, struct InfoField, i);
 
-            g_string_append_printf(output, "%s=%s\n", field.name, field.value);
+            if (field.mark_for_selection || field.tag) {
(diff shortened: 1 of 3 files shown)


hardinfo/info.c
```diff
@@ -56,6 +56,33 @@ void info_add_group(struct Info *info, const gchar *group_name, ...)
...
+    while (1) {
+        struct InfoField field = {
+            .name = va_arg(ap, char *),
+            .value = va_arg(ap, char *),
+            .free_value_on_flatten = TRUE,
+        };
+        if (field.value)
+            field.value = g_strdup(field.value);
+
+        if (!field.name)
```
lpereira: If `!field.name`, `field.value` will leak.

hardinfo/info.c
```diff
@@ -154,10 +181,17 @@ static void flatten_group(GString *output, const struct InfoGroup *group)
...
             field = g_array_index(group->fields, struct InfoField, i);
 
-            g_string_append_printf(output, "%s=%s\n", field.name, field.value);
+            if (field.mark_for_selection || field.tag) {
+                tag = g_strdup_printf("$%s%s$", field.mark_for_selection ? "*" : "", field.tag ? field.tag : "");
+            } else {
+                tag = g_strdup("");
+            }
+
+            g_string_append_printf(output, "%s%s=%s\n", tag, field.name, field.value);
```
lpereira: `tag` isn't freed here.  This might be better to just do something like:

```C
if (field.mark_for_selection) {
        g_string_append_printf(output, "*");
} else if (field.tag) {
        g_string_append_printf(output, with tag);
} else {
      g_string_append_printf(output, without tag...);
}
```

This way you avoid allocating from the heap unnecessarily.

includes/info.h
```diff
@@ -44,17 +44,21 @@ struct InfoGroup {
...
     gchar *value;
 
     int update_interval;
 
     gboolean free_value_on_flatten;
+    gboolean mark_for_selection;
 };
 
 struct Info *info_new(void);
+struct Info *info_new_from_str(const gchar *str); /* opposite of info_flatten(); */
```
lpereira: What would be the use case of this one? I intend to kill `info_flatten()` eventually.

hardinfo/info.c
```diff
@@ -56,6 +56,33 @@ void info_add_group(struct Info *info, const gchar *group_name, ...)
     g_array_append_val(info->groups, group);
 }
 
+void info_add_group_ff(struct Info *info, const gchar *group_name, ...)
```
lpereira: What's `ff` supposed to mean?

includes/info.h
bp0: There's more that's not been pushed yet. info_new_from_str() is needed because I'd started working on a new shell and it uses the info struct exclusively, but so much has not been converted yet. I haven't had any time to work on it this month, I'll try to get back to it.

hardinfo/info.c
bp0: Fast forward. I didn't have a name mind. This pr isn't a real one, it was mostly just for discussion.

bp0: See #180

bp0:
I haven't done anything this in a while, and discussion is at #180. Closing.

PR #182, opened 2017-08-27
dmi: check for null from dmi_get_str() in dmi_chassis_type_str()

bp0:
Fix a crash when there is no DMI available.


lpereira:
Merged, thanks!

PR #185, opened 2017-09-03
bench results: Show Unknown instead of 0 KiB for old results

bp0:
Also, a couple other minor tweaks.


lpereira:
Merged, thanks!

PR #195, opened 2017-09-25
shell: select Computer Summary at start

bp0:


lpereira:
Good idea! Merged, thanks.

Issue #181, opened 2017-08-21
feature: please added info of current display number at screen section


mckaygerhard:
in module computer at screen section all the xorg info are show.. but the only info missing are the most important: the active number of screen.. 

today i try to lauch `Xephyr :1 -ac -screen 900x700 &` and to do i go to my "great hardinfo compilation" and then click on module and ummm what? no screen info are showed!

bp0:
#199 adds the X Display Name to Computer/Display. There is, of course, DISPLAY in environment variables. Is this what you wanted?

mckaygerhard:
hi @bp0 , of course hardinfo wil be run from a desktop environment so let's assumed that DISPLAY environment must be valid and set

very rare cases of limited desktop that are not set, but i cannot remenber right now. so i thinks are good.. please merge

PR #199, opened 2017-10-28
Computer/Display: Add X Display Name

bp0:
A field exists in struct _DisplayInfo, but it wasn't being
filled or shown.

Perhaps fix: https://github.com/lpereira/hardinfo/issues/181


lpereira:
Merged, thanks!

PR #203, opened 2017-11-29
x86: cpu flags fix; show socket:core and thread in proc list

bp0:
* Bug fix: Some flags are just an index, for example power
  management might have "[13] [14]" as flags. This looks like
  a new section to hardinfo shell and it truncates the CPU
  information there.
* Show the Socket:Core and thread in the processor list, for
  x86 only right now. In the future, the idea is to show only
  one line for each core, and list the threads on that core,
  where currently, there is one line for each thread.


lpereira:
Merged, thanks!

PR #205, opened 2017-12-02
x86: add summary of caches to package information

bp0:
Show actual physical caches by counting only unique references from each "cpu" (hardware thread).

Example: Ryzen's 16MB of L3 cache is actually two 8MB caches on each "core complex" inside the processor.
caches


lpereira:
Didn't see this one -- I apologize.  But it's been merged now, thanks!

PR #209, opened 2017-12-11
x86,arm: cpu clocks list in package information

bp0:
Show actual clocks where cores or threads share a clock.
Ex: x86 SMT each core has one clock shared by both threads.
Ex: BCM2837 has one clock for all four cores.
clocks_arm
clocks_x86

### DIFF
--- a/hardinfo/cpu_util.c
+++ b/hardinfo/cpu_util.c
@@ -128,6 +128,11 @@ void cpufreq_update(cpufreq_data *cpufd, int cur_only)
         cpufd->cpukhz_max = get_cpu_int("cpufreq/scaling_max_freq", cpufd->id, 0);
         if (cpufd->scaling_driver == NULL) cpufd->scaling_driver = g_strdup("(Unknown)");
         if (cpufd->scaling_governor == NULL) cpufd->scaling_governor = g_strdup("(Unknown)");
+
+        /* x86 uses freqdomain_cpus, all others use affected_cpus */
+        cpufd->shared_list = get_cpu_str("cpufreq/freqdomain_cpus", cpufd->id);
+        if (cpufd->shared_list == NULL) cpufd->shared_list = get_cpu_str("cpufreq/affected_cpus", cpufd->id);
+        if (cpufd->shared_list == NULL) cpufd->shared_list = g_strdup_printf("%d", cpufd->id);
     }
 }
 
--- a/includes/cpu_util.h
+++ b/includes/cpu_util.h
@@ -25,6 +25,7 @@ typedef struct {
     gint cpukhz_max, cpukhz_min, cpukhz_cur;
     gchar *scaling_driver, *scaling_governor;
     gint transition_latency;
+    gchar *shared_list;
 } cpufreq_data;
 
 typedef struct {
--- a/includes/x86/processor-platform.h
+++ b/includes/x86/processor-platform.h
@@ -31,6 +31,7 @@ struct _ProcessorCache {
     gchar *type;
     gint ways_of_associativity;
     gint uid; /* uid is unique among caches with the same (type, level) */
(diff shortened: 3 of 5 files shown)


bp0:
This PR now also includes a fix for older kernels that didn't provide cache id. 
The oldest tested was 4.9, though. Maybe @mckaygerhard could test older.

lpereira:
This is looking nice.  It would be cool if we could obtain more information on the CPUs -- I never found a practical way.  Tools such as Everest or CPU-Z will offer the number of pins and package type, but they have a hardcoded table (I did some poking around a few years ago).

bp0:
I've thought about this as well. We could create a hard-coded table as well, but it becomes it's own giant project very quickly. I though maybe we could just add a link to cpu-world.com for each processor, but unfortunately there is not an easy way to create such a link with the information we have available.

lpereira:
Yeah, it's kind of a big project. For Intel I know there's
https://ark.intel.com, but I don't think there's any guarantee it'll keep
the links working.  But this is just one manufacturer of CPUs, and wouldn't
solve the issue.

bp0:
There is one interesting piece of information that is available that we aren't showing right now. The CPU socket.
`sudo dmidecode -t 4 | grep Upgrade`
This demands root, though.

mckaygerhard:
i'm ipressive (if the amd 6 core procesor are a normal machine)...i'm quite busy with some php projects, but i can tested at the weekend.. thanks for the support and sorry for late response..

umm about the socket, usefully too by example the Alladin super7 has a buggy AGP port, if user can identify that with hardinfo its easy to provide information interaction..

lpereira:
Take a look here, @bp0: https://github.com/X0rg/CPU-X/blob/master/src/databases.h

We can't use that information directly (that header is licensed under GPLv3 and HardInfo is GPLv2), but it's something to think about.

bp0:
I get the idea. It's certainly possible to have a hard-coded table for that information. Hardinfo already has a few of those, like the Raspberry Pi hardware information, the ARM SOC device tree identifier to name mapping table, the flags definitions, etc. As I was saying, though it become a very large project to maintain on its own. Notice how out-of-date that table is. Perhaps Hardinfo just ends up using libcpuid as CPU-X does.

mckaygerhard:
maybe the technique its like https://github.com/pciutils/pciutils does in apart 3: 3. Getting new IDs

maybe we can made a similar procedure, and updatin info proces to the hardinfo

 The pci.ids file is also mirrored at https://github.com/pciutils/pciids.

we can have a cpu.ids file mirrored at github and the hardinfo can using a update command from hardinfo itseflt exec that update that info of the cpu, inclusivelly the socket and other information!

and thats the solution ot the problem of large project

lpereira:
The server is down at the moment, but HardInfo already updates some of these tables, including the PCI IDs.

mckaygerhard:
well that's the solution to the hardcoded tables... not necesary to use a specific domain/server github can offer raw downloads of the files

lpereira:
I'm not using GitHub for this. While convenient, if they stop existing (like Berlios did), I lose everything again. Hosting a few files somewhere is cheap, and I already pay for hosting services. Eventually the server will be reinstated. Anyway, we're digressing here.

I'll review this patch once I have some time to review them, @bp0.

mckaygerhard:
ok so well a mechanist that test 3 domains, and one of those domain will be github, if down the hardinfo domain if not then other and so..

but the idea its updates the hardcoded table based in offline or online methods

bp0:
I have a branch working on socket information from DMI here: https://github.com/bp0/hardinfo/tree/dmisock but I'd like to see this merged first to avoid a conflict.

mckaygerhard:
hi @bp0 , hi @lpereira sorry for my recently absence i still busy inclusively today due i get these days of "rest" to make many changes at job!!, but happy new year to all of you

PR #212, opened 2017-12-16
Benchmark additions

bp0:
After the bugfix commits were merged as part of #213, this PR has only the two additional benchmarks.
* Al Aburto's flops.c-based, with single-thread/multi-thread variants
* One made up of a couple existing hardinfo benchmark functions, but this time measured by completions in fixed time, and with single-thread/multi-thread variants
* ~~Some results for the these two added to benchmark.conf~~ Moved to #224

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,6 +160,8 @@ set(MODULE_network_SOURCES
 )
 set(MODULE_benchmark_SOURCES_GTKANY
 	modules/benchmark.c
+	modules/benchmark/flopsx.c
+	modules/benchmark/flops_hi.c
 	modules/benchmark/blowfish.c
 	modules/benchmark/cryptohash.c
 	modules/benchmark/drawing.c
@@ -183,6 +185,7 @@ else()
 endif()
 
 set_source_files_properties(
+	modules/benchmark/flopsx.c
 	modules/benchmark/blowfish.c
 	modules/benchmark/fftbench.c
 	modules/benchmark/md5.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,6 +160,7 @@ set(MODULE_network_SOURCES
 )
 set(MODULE_benchmark_SOURCES_GTKANY
 	modules/benchmark.c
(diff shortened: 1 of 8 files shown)


lpereira:
I would prefer if this were split into 4 commits (one for each bullet point).  Makes it easier to bisect later when hunting for a bug.

This benchmark here would be useful as well: https://zsmith.co/bandwidth.html

bp0:
Alright, they're split. One thing we might want to discuss is what should be in the crunch benchmark. I started just using one call to each of the other benchmarks, but settled on just cryptohash and blowfish, because Raspberry Pi 1 could not even complete once in 7 seconds.

Bandwidth bench is interesting.

bp0:
I should point out that the new benchmark just counts how many times in N seconds (I chose 7) that the function cruncher() in crunch.c can be completed. This can be examined here:
https://github.com/lpereira/hardinfo/pull/212/commits/ec050ffb7a22432438c86313330105dd4a99cf15#diff-655ad48be786b17a05534000c4eb2966R30

As I mentioned in the last comment, it is currently just doing one iteration of md5_step, sha1_step, and blowfish encrypt/decrypt. I was hoping to do some fpu stuff in there too, but either fbench() and fft_bench() can't be completed once in 7 seconds on some lower-power systems like Raspberry Pi 1, so it would have to be something new.

bp0:
Added Al Aburto's flops.c benchmark wrapped for hardinfo. It may need a little tweaking, but check it out.

bp0:
I fixed the multi-thread issue in flops.c and it is working well. Here are a few results:
fpu_single
fpu_mt
fpu_mc

bp0:
Funny.
fpu_fun

bp0:
I tried to keep the fixes at the top and the new benchmark commits at the bottom with interactive rebase, but the PR is not following the branch at bp0:newbench, so I've split the fixes into their own PR #213. Those can/should be merged right away, but maybe there should be some discussion about the new benchmarks.

bp0:
I'll make a new PR without flops.c.

PR #214, opened 2018-01-03
Add Intel cpu_insecure bug workaround flag

bp0:
cpu_insecure = page table isolation is being used to work around a security vulnerability in Intel x86 CPUs.

### DIFF
--- a/modules/devices/x86/processor.c
+++ b/modules/devices/x86/processor.c
@@ -341,19 +341,23 @@ gchar *caches_summary(GSList * processors)
     return ret;
 }
 
-
+#define PROC_SCAN_READ_BUFFER_SIZE 896
 GSList *processor_scan(void)
 {
     GSList *procs = NULL, *l = NULL;
     Processor *processor = NULL;
     FILE *cpuinfo;
-    gchar buffer[512];
+    gchar buffer[PROC_SCAN_READ_BUFFER_SIZE];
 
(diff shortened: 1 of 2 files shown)


modules/devices/x86/processor.c
```diff
@@ -341,19 +341,23 @@ gchar *caches_summary(GSList * processors)
...
     FILE *cpuinfo;
-    gchar buffer[512];
+    gchar buffer[PROC_SCAN_READ_BUFFER_SIZE];
 
     cpuinfo = fopen(PROC_CPUINFO, "r");
     if (!cpuinfo)
         return NULL;
 
-    while (fgets(buffer, 512, cpuinfo)) {
+    while (fgets(buffer, PROC_SCAN_READ_BUFFER_SIZE, cpuinfo)) {
```
lpereira: Might be better to use g_file_read_contents() and split on '\n'.  It'll allocate more memory and fragment the heap more, but then we'll not hit these limits.

tsimonq2:
Once this is merged, a backport to the 0.5.1 release (something I can apply to all stable releases of Debian and Ubuntu) would be appreciated.

Thanks for the work on this! :+1:

bp0:
Updated to the flags actually used by 4.15 final. The cpuinfo reading fix could be replaced in the future by rewriting all the processor.c to not use a fixed size buffer.

Issue #217, opened 2018-02-19
Segmentation fault [generating HTML report]


Polesz:
Click generate report, choose filename and run, but

> 
Thread 1 "hardinfo" received signal SIGSEGV, Segmentation fault.
0x00007ffff4c2b036 in __strlen_sse2 () from /lib64/libc.so.6
>

lpereira:
Thanks for the bug report -- are you seeing it with which version? 0.5.1, or something recently built from the current HEAD?

Polesz:
Current HEAD version. OS: Slackware 14.2 (current)

bp0:
Confirmed. The fix from 96980fbf494d9c453c6b558967c90c55e41abe2d needs to be applied to the html  version of that function.

PR #218, opened 2018-02-22
report.c: don't try to use column values that weren't provided

bp0:
(html version)

Like 96980fbf494d9c453c6b558967c90c55e41abe2d, but the html version
of that function. Fixes https://github.com/lpereira/hardinfo/issues/217


lpereira:
Merged, thanks!

PR #222, opened 2018-03-02
report: highlight "selected" items

bp0:
For benchmark results, the current machine will be highlighted in the list by using the shell's select marker.
Text version simply has an asterisk (*) before the highlighted row.
HTML version looks like this now:
hi_report_hl


lpereira:
Merged, thanks!

PR #224, opened 2018-03-05
benchmark result source files and script to build benchmark.conf

bp0:
Should make it easier to manage this information in the future.
If someone wanted to they could create a pull requests for a conf file with their own results and it wouldn't conflict with anything.

Some files include results from the two new benchmarks of #212.


lpereira:
To make maintenance of this data easier, I think it's better to reinstate the server.  I'll see what I can do.

bp0:
I find this handy for now, but I'll just use it myself. Perhaps you could put the server code on github as well.

PR #228, opened 2018-03-12
shell: fix hardinfo window icon

bp0:
Hardinfo no longer has a window icon in the taskbar.
Caused by small typo in 1906d3027dc3ad8b9b5533712299d48e265c852c


lpereira:
Ouch.

PR #229, opened 2018-03-12
benchmark: Add Al Aburto's flops.c

bp0:
Al Aburto's flops.c-based, with single-thread/multi-thread variants.

The original source and readme text are included, but Hardinfo
actually uses a version with the following modifications:
1. All non-Linux code removed
2. Force -DROPT and -DUNIX
3. Multi-thread safe

This file should generate (single thread) output/results identical to
flops.c compiled with -DUNIX -DROPT -O0

The request "NOTE: Please do not remove any of the printouts."
was honored. Hardinfo will send the full printouts to stderr.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -41,6 +41,7 @@ Includes code from or based on:
 * Raytracing benchmark by John Walker (see fbench.c for details)
 * FFT benchmark by Scott Robert Ladd (see fftbench.c for details)
 * Vendor list based on GtkSysInfo by Pissens Sebastien
+* Al Aburto's flops.c
 
 Artwork:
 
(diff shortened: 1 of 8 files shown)


bp0:
I've separated this from the other new benchmark. I really think this one is ready to go.

lpereira:
I don't see any mention of license anywhere in the code. Do you know  what license this thing is licensed under?

bp0:
There is no license specified. 
As it was released, in 1992, as source code, for this purpose, I'm assuming that it is ok. 
What is the license of code with an unspecified license?

lpereira:
If it's unspecified, the only safe bet is to assume it's copyrighted and
proprietary.

bp0:
Alright. It was kindof nice to have a single/multi-thread benchmark though.

Issue #231, opened 2018-03-15
hardinfo -b "non-existing benchmark" causes a crash


bp0:
`hardinfo -b blah`
~~~
** (hardinfo:5178): ERROR **: Unknown benchmark ``blah'' or benchmark.so not loaded
Trace/breakpoint trap
~~~

Oh, that appears to be g_error()'s designed behavior. It calls abort() and terminates the application. I only noticed because now there is the extra message about the trap.

Also from: https://developer.gnome.org/glib/stable/glib-Message-Logging.html#g-error
* "This is not intended for end user error reporting."
* "The message should typically *not* be translated to the user's language."

Is this really the best idea?
I think it would be better to print the message to stderr and set the exit code to non-zero.

PR #232, opened 2018-03-15
Remove use of g_error() for reporting user error

bp0:
Solution to #231.
Instead of g_error(), print message to stderr and set a non-zero exit value.


bp0:
Any problems, questions, or comments?

Issue #225, opened 2018-03-06
Computer/Display information TODO


bp0:
The Computer/Display component of hardinfo is out of date. 

Things to work on:
- [ ] Use `xrandr` to get cards, outputs, and monitors (Don't use GNOME stuff, would fix #200) 
- [ ] Properly interpret core and compatibility profiles for OpenGL and OpenGL ES, and for each card, via `glxinfo` (#27, #170 was an attempt, but I don't think a correct one)
- [ ] Get Vulkan information `vulkaninfo` (fix #171)

mckaygerhard:
umm! this are a "mini" TODO, not a TODO 

^_^u

PR #237, opened 2018-03-18
GPU and Display Information

bp0:
Working on #225, I've started to enumerate GPU devices. Along the way I wrote new PCI utility functions, and changed Devices/PCI to use them. Added X utility functions and changed Computer/Display to use them.

This series also fixes the issue described at #162 for devices/pci.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -215,6 +215,7 @@ endif()
 if (HARDINFO_GTK3)
 add_executable(hardinfo
 	hardinfo/usb_util.c
+	hardinfo/pci_util.c
 	hardinfo/binreloc.c
 	hardinfo/expr.c
         hardinfo/hardinfo.c
@@ -244,6 +245,7 @@ target_link_libraries(hardinfo
 else()
 add_executable(hardinfo
 	hardinfo/usb_util.c
+	hardinfo/pci_util.c
 	hardinfo/binreloc.c
 	hardinfo/expr.c
         hardinfo/hardinfo.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,6 +231,7 @@ add_executable(hardinfo
 	hardinfo/dmi_util.c
 	hardinfo/dt_util.c
 	hardinfo/x_util.c
+	hardinfo/gpu_util.c
 	shell/callbacks.c
 	shell/iconcache.c
 	shell/menu.c
@@ -263,6 +264,7 @@ add_executable(hardinfo
 	hardinfo/dmi_util.c
 	hardinfo/dt_util.c
 	hardinfo/x_util.c
+	hardinfo/gpu_util.c
 	shell/callbacks.c
 	shell/iconcache.c
 	shell/menu.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,6 +143,7 @@ set(MODULE_computer_SOURCES
 set(MODULE_devices_SOURCES
 	modules/devices.c
 	modules/devices/${HARDINFO_ARCH}/processor.c
+	modules/devices/gpu.c
 	modules/devices/battery.c
 	modules/devices/devmemory.c
 	modules/devices/dmi.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,8 @@ if(NOT HARDINFO_NOSYNC)
 	pkg_check_modules(LIBSOUP libsoup-2.4>=2.24)
 endif()
 
+pkg_check_modules(X11 REQUIRED x11)
+
 include(FindZLIB REQUIRED)
 
 include_directories(
@@ -99,10 +101,12 @@ include_directories(
 	${GTK_INCLUDE_DIRS}
 	${LIBSOUP_INCLUDE_DIRS}
 	${ZLIB_INCLUDE_DIRS}
+	${X11_INCLUDE_DIRS}
 )
 link_directories(
 	${GTK_LIBRARY_DIRS}
 	${LIBSOUP_LIBRARY_DIRS}
+	${X11_LIBRARY_DIRS}
 )
 
 set(HARDINFO_MODULES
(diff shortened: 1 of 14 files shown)


mckaygerhard:
puff! you are made a great job!

bp0:
Shots
screenshot_2018-03-17_20-28-46
screenshot_2018-03-17_20-35-28

bp0:
I've squashed some fixes together and re-ordered with `rebase -i` and `push -f`.

includes/computer.h
```diff
@@ -62,14 +62,13 @@ struct _DisplayInfo {
...
-    gchar *extensions;
-    gchar *monitors;
-
     gint width, height;
 
     /* new stuff */
     xrr_info *xrr;
+
+    /* don't free */
+    gchar *vendor; /* points to xrr->xi->vendor */
```
lpereira: If somethings isn't supposed to be freed, it's best to just mark them as `const`.

CMakeLists.txt
```diff
@@ -88,6 +88,8 @@ if(NOT HARDINFO_NOSYNC)
 	pkg_check_modules(LIBSOUP libsoup-2.4>=2.24)
 endif()
 
+pkg_check_modules(X11 REQUIRED x11)
```
lpereira: Is requiring X11 the way to go? GTK3 can use Wayland, and it's not really a good idea to require X11 in that case.

hardinfo/gpu_util.c
```diff
@@ -0,0 +1,218 @@
...
+    return c;
+}
+
+int gpud_list_count(gpud *s) {
+    return gpud_list_append(s, NULL);
+}
+
+/* TODO: In the future, when there is more vendor specific information available in
+ * the gpu struct, then more precise names can be given to each gpu */
+void make_nice_name(gpud *s) {
```
lpereira: Shouldn't this be static?

CMakeLists.txt
mckaygerhard: but older machines need X11 and does not work with Wailand, some distributions does will not include never Wayland and seems for now, if wayland will be used , x11 will be used as base, such as debian 9

hardinfo/gpu_util.c
mckaygerhard: if there a list of names maybe.. but..

mckaygerhard:
umm @bp0 this will sustitue the "Display" section? and will include all the OpenGL/CL/XL/"any-L" information ?

includes/computer.h
bp0: Ok.

hardinfo/gpu_util.c
bp0: Yes, static.

CMakeLists.txt
bp0: It is possible to have a machine with no xserver, nor xrandr, nor xdpyinfo, that still has an X-client application running on it. I was running on a x-less debian machine with `ssh -X ...` and it left the whole display information empty, even though X is clearly being used. The xlib dep is for a fallback of getting information directly from X. Otherwise, (for example, in Debian) we need to make sure that x11-xserver-utils is an install dependency of the hardinfo package, so that we know that xrandr and xdpyinfo will be available.

bp0:
@mckaygerhard, the Display section as it is now is only for information from from X, and if using X over network, it is about actual server being used and not the machine's X server.

There is currently no Wayland information available in Hardinfo.
I've made this clear in this commit:
https://github.com/lpereira/hardinfo/pull/237/commits/19869a588f025e168b29fdd42f2499897017211c

I don't have a setup that can use Wayland, as I have an nVidia card, so I can't really work on it. I do have an old Radeon card that I might try, but it may be too old. To test out detecting multiple cards, I have only one 16x PCIE slot, so I actually used an old PCI S3 Virge card from the 90s that I found in a box. Linux won't even drive it for console anymore, which is sad, but it let me see what two cards in the GPU list would look like.

bp0:
@lpereira, @mckaygerhard: do either of you have a PRIME setup laptop with intel and nvidia graphics?

CMakeLists.txt
bp0: Also, we are guaranteed to have Xlib available if we have GDK. Honestly, we might try and get all the X information directly via Xlib instead of relying on `xrandr` and/or `xdpyinfo`.

includes/computer.h
bp0: Note: making it const introduced a warning when sending it through info_*()

mckaygerhard:
sorry @bp0, not, i have one intel Haswell-ULT with Radeon 8690M, also have some dual intel+radeon very olders firts made.. some time ago i have a intel+nvidia  but i really hate introduce binary only software so i change it

i have access to many hardware but not to a prime: intel+nvidia setup.. sorry, but i can test with my intel+readeon. maybe @nerv55 can help me with that if i ask to

CMakeLists.txt
mckaygerhard: umm so then now X11 its a new depend, so about the reading of scsi older kernels vs newer kernels.. ? may runnign good hardware only work with older kernel..  to the list of TODO (theres at leas one  in the issues) we must added the new depens and runtime-depends and i'll prepare a README introduction update and compilation info how to

bp0:
X11 was always required via gdk, only now the xlib headers are needed to
build. This really doesn't change anything for the binary, it just let's
the cmake magic work to find the right linking config.

CMakeLists.txt
bp0: My reply didn't get added to this thread, but instead the PR thread, so I'll duplicate it here.
X11 was always required via GDK, only now the Xlib headers are needed to build. This really doesn't change anything for the binary, it just lets the cmake magic work to find the right linking config.

bp0:
@mckaygerhard, well I'd certainly appreciate information from any multi-gpu systems. 
Could you send me `glxinfo`, `xrandr --prop`, `dmesg | grep drm` (or complete), `lspci` from any or all of those systems?

bp0:
I've rebased again to simplify the PR.

mckaygerhard:
i'll send you in a couple of hours, let me be free os some task and i feedback here

mckaygerhard:
sorry for late i have a bouild with this pull enable.. i post feedback in later

PR #238, opened 2018-03-19
hardinfo: add cli option to skip benchmarks

bp0:
If you just need the hardware report, this makes it so much faster.


lpereira:
You can skip benchmarks by only loading the computer module with autoload (`hardinfo -m computer.so -a`), but a shortcut like this is easier to remember.

PR #244, opened 2018-03-22
devices/usb: remove old, apparently non-functional, scan methods

bp0:
The sysfs and procfs methods do not appear to work, and the lsusb method was replaced by the lsusb method of usb_util. If a new sysfs method is written, it should be in the framework of usb_util.


mckaygerhard:
seems sysfs wwas also a scan process too, but not working as was happened to me.. but i see in the scan usb method that try from most older to mst newer that was sysfs.. i like that future implemnetations take that in considerations.. this make to work with 2.6.X kernels

mckaygerhard:
2.6 to 3.2 change a lot, minimun to test must be 3.1 i'll merged all in a branch of me and then test all in my branch

lpereira:
HardInfo has been around since the days of kernel 2.4, but I used to run it on machines with a 2.2 kernel... so there are a few scan methods that are long obsolete.

Issue #240, opened 2018-03-20
[bug] empty storage info


mckaygerhard:
the storage module does not show any info, bit with some olders works perfectly

Note that PATA and SATA are not related..

bp0:
What is the content of `cat /proc/scsi/scsi`?

mckaygerhard:
umm reading at the kernels changelog, on recent kernels the /proc/scsi is being "migrated" to sysfs, most distro senable the CONFIG_SCSI_PROC_FS: legacy /proc/scsi/ support to the scsi subsystem, but recently this behaviour is no more..

alternative we can see if something under /sys/class/scsi_host o use the lsscsi program.. in any case we must take in consideration the 2.6 kernel behaviour and the 3.X/4.X kernel new sysfs

```
lsscsi -v
[0:0:0:0]    disk    ATA      ST1000LM024 HN-M 0001  /dev/sda 
  dir: /sys/bus/scsi/devices/0:0:0:0  [/sys/devices/pci0000:00/0000:00:1f.2/ata1/host0/target0:0:0/0:0:0:0]
[1:0:0:0]    cd/dvd  MATSHITA DVD+-RW UJ8E2    1.01  /dev/sr0 
  dir: /sys/bus/scsi/devices/1:0:0:0  [/sys/devices/pci0000:00/0000:00:1f.2/ata2/host1/target1:0:0/1:0:0:0]
# 
cat /proc/scsi/scsi
cat: /proc/scsi/scsi: No existe el fichero o el directorio
```

seems the rute was change.. please if you will upddate, allow searcho on many paths not only newer, take in consideration the machines and older linux that can comlain in those machines.. 

my capture card are not supported in 3.X kernel, framegrabber.. some great hardware like securit camcorders doesn not have modules in recent kernel due "there's no interes" from mantainers
.. 
also many computers are not posible to use with recent kernels that enable PAE by default.. my P3 machines does snot have pae..

bp0:
For now, `lsscsi -c` can be used when `/proc/scsi/scsi` is not found.

bp0:
#245 is a fix, but lsscsi must be installed, and it isn't in debian/ubuntu by default.

mckaygerhard:
as i posted before seems working, but i must test latere in more hardware (implicts to get down to the depot, that's borry! )

so when i'll test report feedback, wheantime i¿'ll close due merged the #245

PR #245, opened 2018-03-25
devices/storage: try `lsscsi -c` if /proc/scsi/scsi is not found

bp0:
This is a hack to address #240. The storage scanner needs work in the future, but this will help for now.


mckaygerhard:
by me, sorry for late,i have try all the pull's related right now please sory for late..

PR #255, opened 2018-05-28
Vendor data file in a new format

bp0:
This is a pci.ids-inspired format for the vendor information. It is more compact and easier to maintain. Multiple match strings can refer to the same vendor.
```
name Advanced Micro Devices
    url www.amd.com
    match_string Advanced Micro Devices
    match_string_case AMD
...
```

Old vendor.conf was based on the .ini format:

```
[vendors]
num=312

[vendor0]
name=Advanced Micro Devices
url=www.amd.com
match_string=Advanced Micro Devices
match_case=0

[vendor1]
name=Advanced Micro Devices
url=www.amd.com
match_string=AMD
match_case=1

[vendor2]
name=Intel
url=www.intel.com
match_string=intel
match_case=0

...
```


lpereira:
Really nice! I might still have a bigger vendor list stashed somewhere, let me see if I can still find it...

lpereira:
@bp0 Found the vendor data, and formatted it in the new format: https://gist.github.com/lpereira/3d60aafb84d4f1902d30d11044708e1b

(Added a new url_support field for those vendors that provide a support URL.)

bp0:
Right on. The `url` line has to be before the `match_string` because the of the simple way it is read. Really only `match_string[_case]` creates an item, simply using the last `name` and `url` it saw. I will do some editing on the file to re-order those lines and and add the `url_support` to the structure. I'm also thinking it would be nice to have `name` and `short_name` or similar.

PR #259, opened 2018-06-03
gpu_util: use vendor_match() in make_nice_name()

bp0:


data/vendor.ids
```diff
@@ -20,26 +22,33 @@ name ASUS
     match_string ASUSTek
     match_string ASUS
 
+# PCI Vendor 1022
```
lpereira: Maybe have a pci_vendor key to match as well?

PR #265, opened 2018-06-09
translation files update

bp0:
hardinfo.pot now has 908 strings (+57)
- [ ] de.po : (706 / 908 remain untranslated)
- [ ] es.po : (225 / 908 remain untranslated)
- [ ] fr.po : (689 / 908 remain untranslated)
- [ ] ru.po : (222 / 908 remain untranslated)


mckaygerhard:
wow, no i must update es.po and other must may also be updated

PR #268, opened 2018-06-21
add zh_CN.po

yetist:
Add Chinese translation.

- [ ] de.po : (706 / 908 remain untranslated)
- [ ] es.po : (225 / 908 remain untranslated)
- [ ] fr.po : (689 / 908 remain untranslated)
- [ ] ru.po : (222 / 908 remain untranslated)
- [ ] zh_CN.po : (211 / 908 remain untranslated)

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: hardinfo\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2018-06-21 17:21+0800\n"
-"PO-Revision-Date: 2018-06-21 16:53+0800\n"
+"POT-Creation-Date: 2018-06-22 16:35+0800\n"
+"PO-Revision-Date: 2018-06-22 16:23+0800\n"
 "Last-Translator: yetist <yetist@gmail.com>\n"
 "Language-Team: Chinese\n"
 "Language: zh_CN\n"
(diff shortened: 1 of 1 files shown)


po/zh_CN.po
```diff
@@ -0,0 +1,5066 @@
...
+
+#: hardinfo/dmi_util.c:144
+msgid "Subnotebook"
+msgstr ""
+
+#: hardinfo/dmi_util.c:145
+#, fuzzy
+msgid "Space-saving"
+msgstr ""
+"强制保持相应的纹理设置为单个的纹理，而不是为了节省空间而由更小的独立纹理组成"
```
lpereira: Is this correct?

po/zh_CN.po
```diff
@@ -0,0 +1,5066 @@
...
+
+#: hardinfo/dmi_util.c:130
+#, fuzzy
+msgid "Invalid chassis type (0)"
+msgstr "%s：无效的邮箱类型"
+
+#: hardinfo/dmi_util.c:131 hardinfo/dmi_util.c:132
+#, fuzzy
+msgid "Unknown chassis type"
+msgstr "unhook：未知钩子类型：%s"
```
lpereira: This doesn't look correct. Why is there a `%s`?

lpereira:
Thanks for the translation! I don't read Chinese, so I can't really review the text; however, some of them do not seem right after passing them through an online automatic translation tool.

mckaygerhard:
@lpereira for those laguajes that are 80% interpreted (pronunciation and intonation are important, which means that it depends on the context) a translation tool cannot detect the context or the intonation.. a light example are russian..

of course some cases has very good trasnlations like spanish in the google translate.. but not same for others.. so well, we must waith for someone that confirms or try to see what wil said @yetist about that

bp0:
There are several strings marked fuzzy, and several that have c-format or access keys that shouldn't. I'd guess that a tool was used, but at least they marked the inexact strings as fuzzy, which means all the non-fuzzy strings are probably acceptable translations.

mckaygerhard:
maybe due seems this file are from a ancient version of hardinfo..

yetist:
Now drop the fuzzy item, and update some item, please review, thank you.
- [ ] zh_CN.po : (527 / 908 remain untranslated)

mckaygerhard:
well @bp0 and @lpereira as i suspect was a translation of and older versino, now updated fuzzy lines gone but at least

yetist:
This is not the old version, because I ran the updatepo.sh script, regenerated the pot file, and translated it based on the new pot file.
As for fuzzy, I use poedit to learn many translation entries, so there will be fuzzy entries.

po/zh_CN.po
bp0: If this is true, it is hilarious.

lpereira:
Merged, thanks!

PR #278, opened 2018-09-08
dt gpu: find gpu in the device tree

bp0:
Will now look for any .../gpu and .../gpu@<address> instead of looking only at /soc/gpu.


lpereira:
Merged, thanks!

Issue #282, opened 2018-09-25
arm/x86: Better handling required where cpufreq information is not available


bp0:
When cpufreq (or cpuinfo clock) information is not available, there can be results like these observed on a RPi3B running Arch.

`Logical CPU Config: 4x 0.00 MHz`

and

```
Clocks 
- 0.00 MHz - 0.00 MHz: 4x
```
where the clock count is also wrong because relationships aren't available.

lpereira:
Closed by #283.

PR #283, opened 2018-09-26
arm/x86: clocks list empty when cpufreq is not available

bp0:
Fixes clock list for #282.


lpereira:
Merged, thanks! (Also, closed #282.)

PR #284, opened 2018-09-27
device tree: some operating-points-v2 properties

bp0:
Decode some properties from `operating-points-v2`, see https://www.kernel.org/doc/Documentation/devicetree/bindings/opp/opp.txt.


lpereira:
Merged, thanks!

PR #286, opened 2018-09-29
 devicetree: get opp-v1/v2 freq range for gpu, if available [v2]

bp0:
* opp = "operating-points", frequency scaling information
  from device tree that can be used for cpu, gpu, etc.
  https://www.kernel.org/doc/Documentation/devicetree/bindings/opp/opp.txt
* adds helper function to get the opp-v1/opp-v2 range of frequencies
  for a node, dtr_get_opp_range() in dt_util.c
* adds a freq range in opp-v2 property for a node in dt
* reports a gpu's max clock frequency if avaiable via opp-v1/v2

### DIFF
--- a/hardinfo/dt_util.c
+++ b/hardinfo/dt_util.c
@@ -942,6 +942,7 @@ dt_opp_range *dtr_get_opp_range(dtr *s, const char *name) {
     uint64_t khz = 0;
     uint32_t lns = 0;
     char *row_status = NULL;
+    uint32_t i = 0;
 
     if (!s)
         return NULL;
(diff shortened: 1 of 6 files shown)


hardinfo/dt_util.c
```diff
@@ -482,6 +483,21 @@ uint32_t dtr_get_prop_u32(dtr *s, dtr_obj *node, const char *name) {
...
+uint64_t dtr_get_prop_u64(dtr *s, dtr_obj *node, const char *name) {
+    dtr_obj *prop;
+    char *ptmp;
+    uint64_t ret = 0;
+
+    ptmp = g_strdup_printf("%s/%s", (node == NULL) ? "" : node->path, name);
+    prop = dtr_obj_read(s, ptmp);
+    if (prop != NULL && prop->data != NULL) {
+        ret = be64toh(*prop->data_int64);
+        dtr_obj_free(prop);
```
lpereira: Is it possible for prop->data to be NULL when prop isn't? If so, this will leak

hardinfo/dt_util.c
```diff
@@ -895,6 +928,86 @@ int dtr_inh_find(dtr_obj *obj, char *qprop, int limit) {
...
+
+    tab_compat = dtr_get_prop_str(s, table_obj, "compatible");
+    tab_status = dtr_get_prop_str(s, table_obj, "status");
+
+    if (!tab_compat || strcmp(tab_compat, "operating-points-v2") != 0)
+        goto get_opp_finish;
+    if (tab_status && strcmp(tab_status, "disabled") == 0)
+        goto get_opp_finish;
+
+    ret = malloc(sizeof(dt_opp_range));
```
lpereira: It's usually not a good idea to mix malloc and g_new in a Glib program; it's too easy to forget to call free() instead of g_free()

hardinfo/dt_util.c
```diff
@@ -546,19 +562,33 @@ char *dtr_elem_phref(dtr *s, dt_uint e, int show_path) {
...
+    dtr_obj *parent = dtr_get_parent_obj(obj);
+    if (parent) {
+        dt_opp_range *opp = dtr_get_opp_range(obj->dt, parent->path);
+        if (opp) {
+            snprintf(opp_str, 511, "[%d - %d %s]", opp->khz_min, opp->khz_max, _("kHz"));
+            free(opp);
+        }
+        dtr_obj_free(parent);
+    }
+    return dtr_elem_phref(obj->dt, *obj->data_int, 1, opp_str);
```
lpereira: I don't know what dtr_elem_phref() does, but does it copy opp_str?

bp0:
dtr_elem_phref() formats the element as a phandle reference. It will copy
opp_str here.

bp0:
.data and .data_int64 are the same pointer in a union.

bp0:
The requested changes have been made.

bp0:
Any problems?

bp0:
Screenshot showing OPPv2:
hi_opp

PR #288, opened 2018-09-30
bench_results: bug fix for cpus with mixed config

bp0:
sscanf() was failing on cpu config strings like: `4x 1459.20 MHz + 4x 1113.60 MHz`. Fixed now.


lpereira:
Merged, thanks!

PR #289, opened 2018-10-01
devices: Remove product-version from string returned by get_motherboard()

bp0:
This reverts a change 2db563687071c099851c59396bdde29a00dba156.

It seems to me that using product version in this way gives
inconsistent values.
The recent benchmark result submitted demonstrates the problem:
`2.0 / X370 SLI PLUS (MS-7A33) (Micro-Star International Co., Ltd.)`

The product version `2.0` is put in front, but without the product
name or product vendor for context, it doesn't really add anything.
The board vendor, which otherwise would be in front, is now in () at
the end.

When the product version is not defined, then the motherboard is
reported as `<board vendor> <board name>`. The only time it is
`<board name> (<board vendor>)` is when product version is defined.

If this change is applied, that same motherboard would be reported
`Micro-Star International Co., Ltd. X370 SLI PLUS (MS-7A33)`
which, to me, makes more sense for get_motherboard(). I have no idea
what "product" this is version 2.0 of.

### DIFF
--- a/data/vendor.ids
+++ b/data/vendor.ids
@@ -16,9 +16,15 @@
 # Match strings should be unique.
 #
 
-name ASUS
+name Lenovo Group
+    name_short Lenovo
+    url www.lenovo.com
+    url_support support.lenovo.com
+    match_string lenovo
+
+name ASUSTeK Computer
+    name_short ASUS
     url www.asus.com
-    # "ASUSTek" is common
     match_string ASUSTek
     match_string ASUS
 
--- a/data/vendor.ids
+++ b/data/vendor.ids
@@ -80,7 +80,9 @@ name Realtek
     url www.realtek.com.tw
     match_string Realtek
 
-name Toshiba
+name Toshiba Corporation
+# styled in all-caps
+    name_short TOSHIBA
     url www.toshiba.com
     match_string Toshiba
 
--- a/data/vendor.ids
+++ b/data/vendor.ids
@@ -172,9 +172,18 @@ name D-Link
     match_string D-Link
 
 name Gigabyte Technology
+    name_short Gigabyte
     url www.gigabyte.com.tw
     match_string Giga-byte
     match_string Gigabyte
+
+name Micro-Star International
+    name_short MSI
+    url www.msi.com
+    url_support www.msi.com/support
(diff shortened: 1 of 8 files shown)


lpereira:
Yeah, so this is hard.  In my case, product_version is `ThinkPad X220`, which makes a lot of sense. At least it makes more sense than `42917J10 LENOVO`.  Maybe, if `product_version` is set, we could use build this string as `${board_vendor} ${board_name} (${product_name})`?

This way, in my computer, this would show up as `LENOVO 42917J10 (ThinkPad X220)`, which makes sense; in the computer for those submitted bencharks, it would show as `Micro-Star International Co., Ltd. X370 SLI PLUS (MS-7A33) (2.0)`, which doesn't make a lot of sense, but it's kind of impossible to always trust firmware vendors to provide this kind of information in a way that makes sense.

lpereira:
On the ThinkPad:

```
bios_vendor = coreboot
bios_version = CBET4000 4.6-635-gd9d351d-dirty
board_vendor = LENOVO
board_version = ThinkPad X220
chassis_vendor = LENOVO
product_family =
product_name = 42917J0
product_version = ThinkPad X220
sys_vendor = LENOVO
```

bp0:
Oh, so `42917J0` is also from the product group. It is falling back because there is no board name.
Maybe instead of falling back separately it could try for the best from the board group and the best from the product/system group and always give `<board> (<system>)`, perhaps with any duplicates removed.

I'm also curious about which values there are coreboot supplied.

bp0:
@lpereira, What do you get on the Thinkpad while using get_motherboard() from the recent commit to this PR?

bp0:
@lpereira have you been able to try this and see if it gives you an acceptable motherboard name on your Thinkpad?

lpereira:
No, not yet. Will try tonight

lpereira:
It gives me "LENOVO 42917J0 ThinkPad X220 (ThinkPad X220)", which is kind of awkward.  Maybe only show the things in parenthesis if they're different than the one before it?

bp0:
Alright, it will now ignore the duplicate version. I believe your motherboard will now be identified as just "LENOVO 42917J0 ThinkPad X220".

PR #291, opened 2018-10-21
DMI Sockets information

bp0:
Requires running as root, but adds CPU and Memory socket information from DMI.
Here's a screengrab:
hi_sockets


lpereira:
Shouldn't this be under DMI?

lpereira:
Thanks for the PR. I'm merging this as-is, but please move it to DMI if you're feeling like it, just so it's more consistent.

PR #292, opened 2018-10-21
Allow DMI on non-x86

bp0:
There's nothing preventing this from running on non-x86. If it is not available it will just say so, like PCI.


lpereira:
DMI is x86-only, isn't it? Maybe it could be renamed to something less specific

bp0:
It seems that some ARM servers that use EFI may also provide DMI.

lpereira:
Merged, thanks!

Issue #266, opened 2018-06-15
Some HTML showing up when asking for text output


danielhoherd:
Using a fresh install (2018-06-14) of Ubuntu 16.04, I see some HTML when requesting text output.

```
hardinfo -r -f text
...snip...
Resources
---------

-I/O Ports-
<tt>0000-0cf7 </tt>		: PCI Bus 0000:00
<tt>  0000-001f </tt>		: dma1
<tt>  0020-0021 </tt>		: pic1
<tt>  0040-0043 </tt>		: timer0
<tt>  0050-0053 </tt>		: timer1
<tt>  0060-0060 </tt>		: keyboard
<tt>  0061-0061 </tt>		: PNP0800:00
<tt>  0064-0064 </tt>		: keyboard
<tt>  0070-0071 </tt>		: rtc0
<tt>  0080-008f </tt>		: dma page reg
<tt>  00a0-00a1 </tt>		: pic2
...snip...
CPU Blowfish
------------

-CPU Blowfish-
<big><b>This Machine</b></big>	2001 MHz	0.764
Intel(R) Celeron(R) M processor         1.50GHz	(null)	26.1876862
PowerPC 740/750 (280.00MHz)	(null)	172.816713

CPU CryptoHash
--------------

-CPU CryptoHash-
<big><b>This Machine</b></big>	2001 MHz	1522.880

CPU Fibonacci
-------------

-CPU Fibonacci-
<big><b>This Machine</b></big>	2001 MHz	1.415
Intel(R) Celeron(R) M processor         1.50GHz	(null)	8.1375674
PowerPC 740/750 (280.00MHz)	(null)	58.07682

CPU N-Queens
------------

-CPU N-Queens-
<big><b>This Machine</b></big>	2001 MHz	0.514

FPU FFT
-------

-FPU FFT-
<big><b>This Machine</b></big>	2001 MHz	0.833

FPU Raytracing
--------------

-FPU Raytracing-
<big><b>This Machine</b></big>	2001 MHz	3.691
Intel(R) Celeron(R) M processor         1.50GHz	(null)	40.8816714
PowerPC 740/750 (280.00MHz)	(null)	161.312647
```

bp0:
This still happens in master.

PR #294, opened 2018-10-21
don't use html in text-only output

bp0:
Fixes https://github.com/lpereira/hardinfo/issues/266

This is also needed so that vendor urls can be links where
appropriate.


lpereira:
Merged, thanks!  (This is good for now, but in the future, when everything is using the struct Info stuff, emphasis can be done automatically.)

PR #295, opened 2018-10-22
shell/report: allow "moreinfo" details to appear in reports; sort benchmark results in reports

### LINKED ISSUE #223: Include "moreinfo" in reports
The moreinfo data should be included in reports, perhaps selectively. Things like the CPU Package information are completely missing from the report because they are in the moreinfo of the "Package Information" item in Devices/Processor.

bp0:
* Fixes #223
* Used by  Devices/Processors/Package Info
* Used by Devices/Device Tree/Summary
* Can be forced for all items with -w

Somewhat related:
* Report benchmark results are now sorted and limited to the 10 nearest results to avoid clutter.
* A new "shell" report format that is very useful for debugging, is not advertised in --help

### DIFF
--- a/shell/report.c
+++ b/shell/report.c
@@ -34,35 +34,37 @@ static FileTypes file_types[] = {
     {NULL, NULL, NULL, NULL}
 };
 
+/* virtual functions */
 void report_header(ReportContext * ctx)
-{
-    ctx->header(ctx);
-}
+{ ctx->header(ctx); }
 
 void report_footer(ReportContext * ctx)
-{
-    ctx->footer(ctx);
-}
+{ ctx->footer(ctx); }
 
 void report_title(ReportContext * ctx, gchar * text)
-{
-    ctx->title(ctx, text);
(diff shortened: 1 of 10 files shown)


shell/report.c
```diff
@@ -34,35 +34,37 @@ static FileTypes file_types[] = {
     {NULL, NULL, NULL, NULL}
 };
 
+/* virtual functions */
 void report_header(ReportContext * ctx)
-{
-    ctx->header(ctx);
-}
+{ ctx->header(ctx); }
```
lpereira: Why the change in indentation?

lpereira:
Nice feature! Will try this in the next few days

shell/report.c
bp0: Those bugged me a little, and I was just trying to make that section more compact. I guess report_header(ctx) is better than using ctx->header(ctx) directly? Because of the double ctx maybe. Idk.

PR #299, opened 2018-10-26
Fixes for #298

bp0:
For #298:

* [x] W: hardinfo implicit-pointer-decl /home/abuild/rpmbuild/BUILD/hardinfo-20181022T194523/hardinfo/dt_util.c:540
* [x] W: hardinfo implicit-pointer-decl /home/abuild/rpmbuild/BUILD/hardinfo-20181022T194523/hardinfo/vendor.c:119
* [x] W: hardinfo implicit-pointer-decl /home/abuild/rpmbuild/BUILD/hardinfo-20181022T194523/modules/benchmark/bench_results.c:127, 208
* [x] W: hardinfo implicit-pointer-decl /home/abuild/rpmbuild/BUILD/hardinfo-20181022T194523/modules/computer/os.c:396
* [x] W: hardinfo implicit-pointer-decl /home/abuild/rpmbuild/BUILD/hardinfo-20181022T194523/modules/network.c:152

I: Program is using uninitialized variables. Note the difference between "is used" and "may be used"
* [x] W: hardinfo uninitialized-variable /home/abuild/rpmbuild/BUILD/hardinfo-20181022T194523/modules/benchmark/zlib.c:28

I: Program returns random data in a function
* [x] E: hardinfo no-return-in-nonvoid-function /home/abuild/rpmbuild/BUILD/hardinfo-20181022T194523/hardinfo/gpu_util.c:37, 405
* [x] E: hardinfo no-return-in-nonvoid-function /home/abuild/rpmbuild/BUILD/hardinfo-20181022T194523/hardinfo/pci_util.c:202, 160
* [x] E: hardinfo no-return-in-nonvoid-function /home/abuild/rpmbuild/BUILD/hardinfo-20181022T194523/modules/benchmark/bench_results.c:137
* [x] E: hardinfo no-return-in-nonvoid-function /home/abuild/rpmbuild/BUILD/hardinfo-20181022T194523/modules/devices/spd-decode.c:982

----

* [x] E: hardinfo no-return-in-nonvoid-function /home/abuild/rpmbuild/BUILD/hardinfo-20181022T194523/hardinfo/gpu_util.c:405
* [x] E: hardinfo no-return-in-nonvoid-function /home/abuild/rpmbuild/BUILD/hardinfo-20181022T194523/hardinfo/pci_util.c:160

PR #300, opened 2018-10-27
benchmark/zlib: new version of zlib benchmark

bp0:
The new version uses a fixed time (7s) and the results are much
more consistent than the previous version.

A couple results for the new version, one "low" and one "high",
replace the old results, which were quite strange.

This will probably conflict with #299.


bp0:
I've rebased with master and resolved the conflict with #299.

bp0:
@lpereira, why not merge?

lpereira:
Merged. I completely forgot about this one.

PR #302, opened 2018-11-03
devices/sensors - hwmon: Added support for sysfs labels

blacksvk:
If sensor label is not defined in config file, it will be loaded from sysfs


lpereira:
Merged, thanks!

PR #303, opened 2018-11-04
devices/sensors: bug fix - read_sensor_labels never gets called

blacksvk:
That was the reason why sensor labels were not loaded from config file. Fixed now.


lpereira:
Merged, thanks

Issue #304, opened 2018-11-04
Network interface information doesn't show icons


lpereira:
Icons aren't being shown any longer for any of the information screens, including network interface, storage, input, etc.

bp0:
git bisect points to 9eaf1a7a96aabc021594bcc6f5fe2ef60da8fb0c

PR #305, opened 2018-11-05
Fix missing icons

bp0:
Fixes https://github.com/lpereira/hardinfo/issues/304

After 9eaf1a7a96aabc021594bcc6f5fe2ef60da8fb0c, the
whole key part, including the opening and closing '$',
is stored in INFO_TREE_COL_DATA so that the
key_*() functions can be used against it anywhere.

The first '$' no longer needs to be skipped.


lpereira:
Merged, thanks

PR #308, opened 2018-11-09
devices/usb: Add USB interfaces and device icons

ocerman:
- Usb interfaces are shown in more info box
- Device icons are determined based on usb interface class/subclass/protocol and usb device class.


lpereira:
Merged, thanks!

PR #311, opened 2018-11-10
devices/sensors - hwmon: FIX - Ambiguous hash table keys

ocerman:
The hash table is currently using keys based on sensor name, like "temp1". This causes problems when they are multiple drivers that share same sensor name. So for example if a label is defined in config file for driverA/temp1, the same label is incorrectly used also for driverB/temp1.

This PR adds driver name to the hash key and that fixes mentioned issue.


lpereira:
Merged, thanks

PR #314, opened 2018-11-18
Re-added Desktop Environment field and added MATE version detection

ocerman:
The Desktop Environment field was missing in current git master. I think it was removed by mistake, because the `detect_desktop_environment` function seems to be working just fine. I have also added function for detecting MATE version.


lpereira:
Merged, thanks!

PR #315, opened 2018-11-28
devices/storage: added udisks2 support

ocerman:
This PR adds initial udisks2 support. Udisks2 is preinstalled on many popular linux distros and can be used for getting additional drive information.

So far this code is using udisks2 for loading additional SCSI drives information (like serial number, size, rotation rate, smart info...) for Devices/Storage, but it can be later expanded for stuff like partitions info, temperature monitoring in Devices/Sensors, ...

snimka obrazovky 2018-11-28 00-38-27


lpereira:
Merged, thanks!

PR #316, opened 2018-11-28
computer/os: fix parse_os_release

ocerman:
Seems like there was small mistake in for loop. Should be fixed now.

### DIFF
--- a/modules/computer/os.c
+++ b/modules/computer/os.c
@@ -327,7 +327,7 @@ static gchar *
 parse_os_release(void)
 {
     gchar *pretty_name = NULL;
-    gchar **split, *contents, *line;
+    gchar **split, *contents, **line;
 
     if (!g_file_get_contents("/usr/lib/os-release", &contents, NULL, NULL))
         return NULL;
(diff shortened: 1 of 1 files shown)


modules/computer/os.c
```diff
@@ -336,9 +336,9 @@ parse_os_release(void)
     if (!split)
         return NULL;
 
-    for (line = *split; *line; line++) {
-        if (!strncmp(line, "PRETTY_NAME=", sizeof("PRETTY_NAME=") - 1)) {
-            pretty_name = g_strdup(line +
+    for (line = split; *line; line++) {
```
lpereira: I can't see the problem with the original code; what is this patch trying to solve?

ocerman: So the parse_os_release did not worked for me. I figured out that the for-loop is not actually looping the lines. When you add `printf("%s \n", line)` into for-loop you can see what it is actually doing:
```
ondrej@ryzen3:~/hardinfo/build$ ./hardinfo 
NAME="Ubuntu" 
AME="Ubuntu" 
ME="Ubuntu" 
E="Ubuntu" 
="Ubuntu" 
"Ubuntu" 
Ubuntu" 
buntu" 
untu" 
ntu" 
tu" 
u" 
" 
```
After this PR it seems to be working fine:
```
ondrej@ryzen3:~/hardinfo/build$ ./hardinfo 
NAME="Ubuntu" 
VERSION="18.10 (Cosmic Cuttlefish)" 
ID=ubuntu 
ID_LIKE=debian 
PRETTY_NAME="Ubuntu 18.10" 
```

lpereira: Ah, indeed, you're right! Thanks for the fix

PR #324, opened 2018-12-04
Replaces the uber-graph submodule with a local fork

bp0:
Includes the hardinfo-required source files directly instead of using a git submodule. The original is unmaintained. Uber-graph is only used with the experimental GTK3 build.

@tsimonq2 : Does this help with https://github.com/lpereira/hardinfo/issues/101#issuecomment-329985085 ?


mckaygerhard:
no matter what it comment means, if now it is forgotten and unmantained the sources of that sub-module can be treated as part of the project sources!

Issue #312, opened 2018-11-10
Sensor monitoring function is causing enough load to rise Core voltage


ocerman:
I tried to debug why is hardinfo reporting much higher Core voltage than `lm-sensors`. I discovered that the reported number is technically correct, but very confusing for end user.

What happends is that hardinfo's monitoring function causes enough load for the idle system to rise the cpu voltage (propably also the clocks) for the fraction of the second.

I found the workaround: adding `usleep(1)` just before `read_raw_hwmon_value` fixed my problem, even though this probably is not the best solution.

I'm opening this issue hoping someone else may have better idea how to fix this? Will it be OK if I open PR with usleep workaround?

bp0:
I think so. I did try it myself and didn't notice a change.

blacksvk:
It seems like not every sensor can update their value fast enough to sense this.
I got this issue on my Ryzen 3 system with SVI2 Core Voltage sensor (with `ondemand` governor). Motherboard sensors are not affected

bp0:
I have a Ryzen system and it only has monitoring via it8655. I did not notice any difference with and without the delay. The delay doesn't really cause any problems, and if adding it improves the situation for your case, then I think it could be included.

ocerman:
Fixed by #327

PR #327, opened 2018-12-09
devices/sensors: hwmon improvements

ocerman:
- Performance improvement: instead of blindly trying to read sensor files 256 times, the sensor file names are determined based on directory listing. Because `g_dir_read_name` do not sort filenames, it is first determined which files exists and then read in for-loop. This also fixes my problems described in issue #312 
 - Added support for Current and Power sensors
 - Removed unwanted whitespace from sensor labels


lpereira:
Merged, thanks -- the usleep(1) thing looked too unreliable; this is way better.

PR #330, opened 2019-01-12
Udisks2 as primary source for storage info + added drive temperatures to sensors view

ocerman:
This PR modifies udisks2 code so instead of using udisks2 only for loading additional information for detected SCSI drives, we can use udisks2 as standalone source for all storage devices.

This brings 2 benefits:
- Support for new storage devices (NVMe and floppy drives)
- Support for Linux distros which do not provide `/proc/scsi` nor `lssci` by default (for example BunsenLabs Linux)

To avoid drives being detected twice (by udisks2 and by other function), functions `__scan_scsi_devices` and `__scan_ide_devices` are called only when udisks2 is not able to detect any drives.

This PR also adds drive temperatures (from udisks2) to sensors view.


lpereira:
Thank you!

hardinfo/udisks2_util.c
```diff
@@ -229,6 +240,41 @@ void udiskd_free(udiskd *u) {
...
+    }
+
+    if (!smart_enabled) {
+        return NULL;
+    }
+
+    v = get_dbus_property(drive, UDISKS2_DRIVE_ATA_INTERFACE, "SmartTemperature");
+    if (v) {
+        disk_temp = udiskt_new();
+        disk_temp->temperature = (gint32) g_variant_get_double(v) - 273.15;
```
kmikita, line 261: #634

PR #332, opened 2019-01-16
vendor.ids: newer version

bp0:
...from https://github.com/bp0/verbose-spork/blob/master/data/vendor.ids
Includes the two recently added plus some more. 
ansi_color and match_string_exact will just be ignored by hardinfo, which doesn't use them
yet.


lpereira:
Nice!

PR #335, opened 2019-05-20
Fix: Memory SPD has the wrong viewtype if there is an error

bp0:
First commit is format cleanup, second commit is the fix.


lpereira:
Thanks! Merged!

PR #336, opened 2019-05-23
devices/gpu: amdgpu - core and memory clocks

ocerman:
Show the min/max core and memory frequency for AMD GPUs with amdgpu driver

### DIFF
--- a/hardinfo/gpu_util.c
+++ b/hardinfo/gpu_util.c
@@ -133,8 +133,10 @@ static void amdgpu_parse_dpmclk(gchar *path, int *min, int *max) {
 
 static void amdgpu_fill_freq(gpud *s) {
     gchar path[256] = "";
-    int clk_min = -1, clk_max = -1;
+    int clk_min = -1, clk_max = -1, mem_clk_min = -1, mem_clk_max = -1;
+
     if (s->sysfs_drm_path) {
+        /* core */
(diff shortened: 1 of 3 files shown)


hardinfo/gpu_util.c
```diff
@@ -101,6 +101,61 @@ static void intel_fill_freq(gpud *s) {
...
+    int sp, i, clk;
+
+    *min = -1;
+    *max = -1;
+
+    g_file_get_contents(path, &data, NULL, NULL);
+    if (data) {
+        p = data;
+
+        while(next_nl = strchr(p, '\n')) {
```
lpereira: Won't GCC whine here because it thinks you should use `==` instead of `=`? (It stops whining if you put an extra set of parenthesis around the `while` condition.)

hardinfo/gpu_util.c
```diff
@@ -101,6 +101,61 @@ static void intel_fill_freq(gpud *s) {
...
+            }
+
+            p = next_nl + 1;
+        }
+    }
+    g_free(data);
+}
+
+static void amdgpu_fill_freq(gpud *s) {
+    gchar path[256] = "";
```
lpereira: Might be better to use `PATH_MAX` just in case.

lpereira:
Merged, thanks!

PR #337, opened 2019-05-25
devices/sensors - Sensors.conf parsing fixes

ocerman:
1. commit - Formulas in a compute statement in sensors.conf are separated with comma. Current implementation worked only when there was space before comma. This commit should fix that.

 2. commit - Driver name was so far used as:
- part of label in "Sensor" column,
- part of the moreinfo key,
- label which was used to search actual hwmon device in sensors.conf (by "chip" statement). This was the issue, because according to the sensors.conf man page: chip statement contains "chip type"; which seems to be whatever is written in "name" file (according to sensors app output). Driver name and device name ("name" file) are not always the same. That's why I replaced driver name with device name.

### DIFF
--- a/modules/devices/sensors.c
+++ b/modules/devices/sensors.c
@@ -29,7 +29,7 @@ GHashTable *sensor_compute = NULL;
 GHashTable *sensor_labels = NULL;
 gboolean hwmon_first_run = TRUE;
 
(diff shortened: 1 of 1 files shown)


lpereira:
Thanks! Merged :)

PR #339, opened 2019-05-26
devices/spd: Improved DDR4 SPD support

ocerman:
- Added Voltage
- Added initial XMP 2.0 support
- Fixed DDR4 Manufacturer


lpereira:
Merged, thanks!

PR #341, opened 2019-06-02
devices/spd: More DDR4 improvements

ocerman:
Commit 1: Added manufacturing date
Commit 2: Added DRAM Manufacturer
Commit 3: FIX - Partial data is evaluated every refresh (not only on first time)


bp0:
https://github.com/lpereira/hardinfo/blob/8c4845a39d030a78422c5d2d7cbfdce91e40615a/modules/devices/spd-decode.c#L1685

That needs to be set to NULL to prevent freeing uninitialized pointer when dimm_list is empty.

I noticed this when loading the ee1004 driver on a machine that, turns out, has only ddr3. It might be a good idea to check for any ee1004 devices and then fallback to using eeprom if none are found.

lpereira:
Merged, thanks!

PR #343, opened 2019-06-02
devices/spd-decode: Fallback to another driver when no devices are found

ocerman:
This PR implements @bp0's idea (in #341) to fallback to another driver when no devices are found. (So when no ee1004 devices are found, hardinfo will fallback to eeprom driver)

This also fixes freeing uninitialized pointer problem in decode_dimms function (@bp0 reported that in #341).


lpereira:
Merged, thanks!

Issue #345, opened 2019-06-16
DDR3 SPD decode, reported size is double actual size for Samsung parts


bp0:
As seen below, Memory SPD reports 4096MB for a 2048MB DIMM. The DMI information is correct.
![frog-dmi](https://user-images.githubusercontent.com/1758090/59566169-cbdba200-9021-11e9-8efb-3df114b94493.png)
![frog-spd](https://user-images.githubusercontent.com/1758090/59566173-ced69280-9021-11e9-8226-336140083e33.png)
![frog-eeprom](https://user-images.githubusercontent.com/1758090/59566176-dc8c1800-9021-11e9-8816-cc5b53c44268.png)

bp0:
Here is a system with one Samsung and one Hynix DIMM. Only the Samsung DIMM has a reported size that is double the actual size.

![leene-dmi](https://user-images.githubusercontent.com/1758090/59566299-e6167f80-9023-11e9-9957-6d502b5d78d1.png)
![leene-spd](https://user-images.githubusercontent.com/1758090/59566303-ef075100-9023-11e9-8c4f-b77b4e34669e.png)
![leene-spd-hynix](https://user-images.githubusercontent.com/1758090/59566307-fb8ba980-9023-11e9-9284-43e5345960f9.png)
![leene-spd-samsung](https://user-images.githubusercontent.com/1758090/59566308-fd556d00-9023-11e9-85f9-e485d8e4a683.png)

lpereira:
In other news, it would be nice to not show the DIMM information in the CPU information, and move it to the Memory SPD thing.

bp0:
Yes, I would like that too, but I think the problem I had is that I don't know a way to match the information.

ocerman:
I think we could move DMI DIMM information into Devices->DMI, as matching it to SPD would be difficult.

Or,  we could move current /proc/meminfo from Devices->Memory to Computer->Memory which I think would be better place, because values from /proc/meminfo is mostly dependent on OS and its usage. Then we could move DMI DIMM information into Devices->Memory.

ocerman:
Also, about original issue: Timings are probably also not correct (tCL should not be 0). Also I noticed that decoding Module Manufacturer is not implemented for DDR3. I may take a look on it.

PR #354, opened 2019-06-20
move devices/memory to computer/memory usage

bp0:
As discussed in https://github.com/lpereira/hardinfo/issues/345.

Everything seems to be working, except there is now a message at startup:
> Module "Computer" depends on module "devices.so"

Didn't it always? I'm not sure what I changed that causes this message. This also blocks benchmarks from working, because the spawned process dies with this message.
Maybe @lpereira could fix that?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,6 +135,7 @@ set(MODULE_computer_SOURCES
 	modules/computer/languages.c
 	modules/computer/loadavg.c
 	modules/computer/memory.c
+	modules/computer/memory_usage.c
 	modules/computer/modules.c
 	modules/computer/os.c
 	modules/computer/uptime.c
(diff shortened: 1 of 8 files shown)


lpereira:
Making various modules was a mistake.  I've considered rewriting HardInfo a million times and getting rid of this mess is one of those things that makes the top 5 reasons to do so.  (The shell communicating with the modules via a .ini-style string is number 1 though.)

I'll take a look, see how hard it is to fix this warning/error.

lpereira:
Ah, the reason is that the `devices.so` module requires the `memlabels` GHashTable, but it's defined in `computer.so`.  Just pushed a commit that debug-prints the value of `g_module_error()` when loading a module.

bp0:
Alright, should be fixed.

lpereira:
Merged, thanks!

PR #355, opened 2019-06-21
DMI socket/module information moved from Processor to Memory DMI

bp0:
As discussed #345. Following #354.

SPD is still preferable because Hardinfo doesn't need to run as root, only the user has to add the required module. I would still like to merge Memory DMI and Memory SPD in the future, if a way can be found to match the information from DMI and SPD and create an abstract Memory Slot/Module structure to display.


lpereira:
Thanks, merged!

PR #356, opened 2019-06-21
Not(e) for missing nvme info without udisks2

bp0:
As @ocerman mentioned recently in #243. 
`/sys/class/nvme` could be used, but is there any recent system that doesn't include udisks2 that also might have the relatively new nvme?

edit: note a slight conflict with #355 that should be easy to resolve.

### DIFF
--- a/modules/devices.c
+++ b/modules/devices.c
@@ -115,6 +115,7 @@ gchar *printer_list = NULL;
 gchar *printer_icons = NULL;
 gchar *pci_list = NULL;
 gchar *input_list = NULL;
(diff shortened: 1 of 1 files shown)


ocerman:
I really do not know any current Linux distro that doesn't include udisks2.

Also... udisksctl is not really needed, only udisksd.

bp0:
Thanks, I've updated the note.
Maybe systemd-less systems?

lpereira:
I agree that a note like this is needed. Merging.

PR #360, opened 2019-06-23
Change "smart" to "S.M.A.R.T."

TotalCaesar659:


bp0:
Maybe just "S.M.A.R.T" or "Self-monitoring" instead of ""S.M.A.R.T monitoring" because the M already stands for monitoring. It's redundant, like "ATM Machine".

TotalCaesar659:
Well, I just wanted to make it more clearer, cause S.M.A.R.T. is already well-established. And I have a doubt that some tautology will play a big role (especially in Linux world with its "GNU is not Unix" etc).
Let's see what @lpereira will say about it, of course I may change the PR.

bp0:
Well I agree that Smart -> S.M.A.R.T is a good idea, but I suggest as a further refinement that we also remove the redundant term in some way. "Self-monitoring (S.M.A.R.T)" might be another candidate.

As "ATM Machine" = "Automatic Teller **Machine Machine**", a common error, so does "S.M.A.R.T Monitoring" = "Self-**Monitoring**, Analysis and Reporting Technology **Monitoring**". There is a redundant term.

modules/devices/storage.c
```diff
@@ -86,7 +86,7 @@ gboolean __scan_udisks2_devices(void) {
             features = h_strdup_cprintf(", %s", features, _("Ejectable"));
         }
         if (disk->smart_supported) {
-            features = h_strdup_cprintf(", %s", features, _("Smart monitoring"));
+            features = h_strdup_cprintf(", %s", features, _("S.M.A.R.T. monitoring"));
```
lpereira:
```suggestion
            features = h_strdup_cprintf(", %s", features, _("Self-monitoring (S.M.A.R.T.)"));
```

modules/devices/storage.c
```diff
@@ -139,7 +139,7 @@ gboolean __scan_udisks2_devices(void) {
             moreinfo = h_strdup_cprintf(_("Connection bus=%s\n"), moreinfo, disk->connection_bus);
         }
         if (disk->smart_enabled) {
-            moreinfo = h_strdup_cprintf(_("[Smart monitoring]\n"
+            moreinfo = h_strdup_cprintf(_("[S.M.A.R.T. monitoring]\n"
```
lpereira:
```suggestion
            moreinfo = h_strdup_cprintf(_("[Self-monitoring (S.M.A.R.T.)]\n"
```

TotalCaesar659:
@lpereira Updated.

lpereira:
Merged, thanks!

PR #366, opened 2019-06-26
devices/dmi_memory.c: shell viewtype=1 version, other tweaks

bp0:
ViewType=1 version:
dmimem

Notes:
* add -std=c99 for dmi_memory.c to use a unicode arrow
* -std=c99 is picky about inline functions


lpereira:
Nice!

PR #334, opened 2019-05-13
devices/spd: Add initial DDR4 SPD support

ocerman:
Snímka obrazovky 2019-05-13 23-30-34

This PR adds initial support for decoding DDR4 SPD. It is needed to have ee1004 driver loaded and properly configured.

If you are getting hinote _A current driver has provided only partial DDR4 SPD data_:
 - make sure ee1004 driver is loaded
 - unload eeprom driver
 - Follow these instructions how to instantiate ee1004 devices manually  (https://www.spinics.net/lists/linux-i2c/msg32331.html):

> First you must find out the i2c bus number of your motherboard's SMBus,
> using i2cdetect (i2cdetect is part of the i2c-tools package.) Example:
> ```
> # modprobe i2c-dev
> # i2cdetect -l | grep smbus
> i2c-4	smbus     	SMBus I801 adapter at f040      	SMBus adapter
> ```
> So the SMBus is i2c-4. Then you must find the I2C addresses of the
> SPD EEPROMs:
> ```
> # i2cdetect -y 4 0x50 0x57
>      0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
> 00:                                                 
> 10:                                                 
> 20:                                                 
> 30:                                                 
> 40:                                                 
> 50: -- 51 -- 53 -- -- -- --                         
> 60:                                                 
> 70:                                                 
> ```
> So there are 2 DDR4 SPD EEPROMs, one at I2C address 0x51 and one at
> I2C address 0x53. To instantiate the devices, do:
> ```
> # echo ee1004 0x51 > /sys/bus/i2c/devices/i2c-4/new_device 
> # echo ee1004 0x53 > /sys/bus/i2c/devices/i2c-4/new_device 
> ```

PR #369, opened 2019-06-27
[WIP; Don't Merge] Unify DMI and SPD memory information into one "Memory Devices"

bp0:


bp0:
First with no information, second with only SPD, third with both DMI and SPD. Should work with DDR2,3,4. spd-decode.c is a bit of a mess to try and reform.
Screenshot from 2019-06-27 02-05-32
Screenshot from 2019-06-27 02-06-46
Screenshot from 2019-06-25 18-42-35

lpereira:
Looking good!

bp0:
It seems in pretty good shape, but I can't test very much. I have several DDR3 machines, and one DDR4 machine, but the ee1004 driver doesn't work on the DDR4 machine for some reason.
@lpereira or @ocerman , if you have some DDR4 machines could you test it out and maybe even post some screengrabs or the "Memory Devices" section of `sudo hardinfo -r -f text -s -w`?

lpereira:
[Some screengrabs from my DDR4 machine are here](https://imgur.com/a/uc62SYo).  Unfortunately the ee1004 module doesn't give out any information here either.

bp0:
Thanks. Do you get any extra information with the regular eeprom module?

bp0:
@lpereira , it seems that there is no auto-detect of devices yet. Also, eeprom and ee1004 will compete so they must not be loaded together.
The process to configure the ee1004 devices is described at https://www.spinics.net/lists/linux-i2c/msg32331.html
After following this guide, I was able to see the DDR4 SPD information, but this is not a great experience yet.

bp0:
Of course, @ocerman  said all of that in his original PR #334. Sorry for my ignorance.

PR #370, opened 2019-06-27
GTK3/Benchmarks disable guibench only

jurobystricky:
GTK3 build inadvertably disabled all benchmarking.
At present, the GTK3 build intentionally avoids building of "guibench"
due to non-trivial changes from GTK2 to GTK3.
However, all other benchmarks were disabled as well.
This was not by design, as non-gui benchmarks have no GTK3
dependencies.
The reason benchmarks were diasbled (did not show up in the menus)
was that the code for "benchark_gui" still referenced the missing "guibench"
routine. This routine was not present in the "benchmark.so", so the library
would not load due to the missing entry point.

This patch addresses the issue. All benchmarks (except guibench)
show/work fine with GTK3.

Signed-off-by: Juro Bystricky <juro.bystricky@intel.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,7 +170,6 @@ set(MODULE_benchmark_SOURCES_GTKANY
 	modules/benchmark/blowfish.c
 	modules/benchmark/blowfish2.c
 	modules/benchmark/cryptohash.c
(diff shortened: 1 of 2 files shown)


lpereira:
Thanks, merged!

Issue #386, opened 2019-06-30
DMI Memory location conflict


ocerman:
First of all, thanks @bp0 for Unifying DMI/SPD, it is much better experience now.

I have one issue in my box. My BIOS is reporting DIMM numbers for each channel separately. So I see 2x DIMM 0 and 2x DIMM1 in the list. This is (propably) also causing the issue that same "more info" is loaded for both sticks (look at DMI handle).

I suggest we could change Locator string by adding "Bank locator". So on my system it would be like Mainboard > P0 Channel A > DIMM 0.

DMI decode -t memory output:  [dmimem.txt](https://github.com/lpereira/hardinfo/files/3342870/dmimem.txt)
Hardinfo report: [hadinfomemdev.txt](https://github.com/lpereira/hardinfo/files/3342871/hadinfomemdev.txt) (also sorry @bp0, i have just noticed that you were asking for this in #369 )

bp0:
Hmm something else is wrong too: the SPD vendor should have been pulled up in to the DMI vendor field to replace the "Unknown"s. ([Code for that](https://github.com/lpereira/hardinfo/blob/master/modules/devices/dmi_memory.c#L301,L304))

Anyway, about your original issue, that is a good suggestion. Although, to try and keep it short I think I will first only including Bank Locator when there are duplicate locators. I already shortened "System Board Or Motherboard" to just "Mainboard".

What is the model of motherboard if I may ask?

ocerman:
It is ASRock B350 Pro4

As for the Unknown manufacturer: It seems that the problem is that there is [no check](https://github.com/lpereira/hardinfo/blob/master/modules/devices/dmi_memory.c#L237) for "Unknown", only for "\<BAD INDEX\>"

bp0:
I have an ASUS PRIME B350 PLUS, which seems to be the ASUS version of that board, and mine are:
```
Mainboard > BANK 0 > DIMM_A1
Mainboard > BANK 1 > DIMM_A2
Mainboard > BANK 2 > DIMM_B1
Mainboard > BANK 3 > DIMM_B2
```
So Bank Locator doesn't really tell me anything.

bp0:
Also I have an HP Board with:
```
Mainboard > BANK 0 > DIMM_1
Mainboard > BANK 1 > DIMM_2
Mainboard > BANK 2 > DIMM_3
Mainboard > BANK 3 > DIMM_4
```
Bank Locator again seems meaningless.

[...]

PR #387, opened 2019-06-30
Memory devices: show bank in locator if needed 

bp0:
If a locator of array > socket is not unique, then use array > bank > socket in the list.
As suggested in #386.

How does this work for you @ocerman ?


ocerman:
@bp0 works fine, thanks

mem

bp0:
@ocerman, what do you think about https://github.com/lpereira/hardinfo/pull/387/commits/705081b3125bc013859a1d73229d14f0cc429010?

bp0:
@lpereira, this should be ready to merge.

PR #393, opened 2019-07-02
struct info: add moreinfo() tag and flags support, field sort

### LINKED ISSUE #391: Key flag support is missing in info() and icon handling
A while ago I added support for flags in the key part of the shell format:
* `*` (asterisk) marks the item to be selected/highlighted (use most notably in bench results)
* `!` (exclaimation) tells the report to include the moreinfo()

The `key` in this case is the whole left side of the key=value pair in the shell format.
The syntax would be:
`[$[!][*][<tag>]$]<name>=<value>`

I also added some helper functions to use them (in shell.c):
```
gboolean key_is_flagged(gchar *key); /* the key has a $...$ section */
gboolean key_is_highlighted(gchar *key); /* wants to be highlighted/selected "*" */
gboolean key_wants_details(gchar *key); /* include details in reports "!" */
gchar *key_mi_tag(gchar *key); /* get the moreinfo/icon lookup tag */
const gchar *key_get_name(gchar *key); /* get the visible name of the row */
```

Currently using any flag messes up icons. Also, the info struct stuff has no support for them yet.

bp0:
Fixes #391.

### DIFF
--- a/hardinfo/info.c
+++ b/hardinfo/info.c
@@ -40,7 +40,7 @@ void info_group_add_fieldsv(struct InfoGroup *group, va_list ap)
     while (1) {
         struct InfoField field = va_arg(ap, struct InfoField);
 
-        if (field.magic == INFO_LAST_MARKER)
+        if (!field.name)
             break;
         g_array_append_val(group->fields, field);
     }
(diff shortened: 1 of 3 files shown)


lpereira:
I wonder if we should fix the Info API so that we can write it like so:

```C
info_add_group("Group Name",
               info_field(.key="Some Key", .value="Some Value"),
               info_field(.key="Another Key", .icon="icon.png"),
               info_field());
```

With a macro, this is possible:

```C
#define info_field(...) (struct InfoField) { \
    .some_member_with_default_value = 3, \
    __VA_ARGS__ \
}
```

In this case, even if `info_field()` is called as `info_field(.some_member_with_default_value=50)`, it'll still be initialized with the correct value, superseding the one set in the macro.

lpereira:
Would be nice if we could implement such macro for `info_field_printf()`, too.  Maybe...

```C
#define info_field_printf(fmt, printf_args, ...) (struct InfoField) { \
    .flags = INFO_FIELD_DYNAMIC_VALUE, \
    .value = g_strdup_printf(fmt, printf_args), \
    __VA_LIST__ \
}
```

However this doesn't work, as `printf_args` would have to be in parenthesis (so it's not expanded inside the `InfoField` struct), but can't have parenthesis when `g_strdup_printf()` is being called.

Any thoughts?

bp0:
I like this. I implemented the macro in your first comment in a way that doesn't require changing any existing usage. For the second one (info_field_printf) I would think that the second option is the best so that .flags doesn't conflict with a supplied .flags.

hardinfo/info.c
```diff
@@ -40,7 +40,7 @@ void info_group_add_fieldsv(struct InfoGroup *group, va_list ap)
     while (1) {
         struct InfoField field = va_arg(ap, struct InfoField);
 
-        if (!field.name)
```
lpereira: Any reason to add a magic field and not rely on field.name being NULL?

includes/info.h
```diff
@@ -21,6 +21,13 @@
 #include <stdarg.h>
 #include <glib.h>
 
+enum {
```
lpereira: Maybe flags should be individual booleans -- although less efficient, it's easier to specify them in the macros individually.

hardinfo/info.c
bp0: I was just looking for a use for that .some_member_with_default_value, it was just an idea.

includes/info.h
bp0: Yeah that's probably right.

bp0:
As it is, it always generates a tag if one wasn't provided, where before it only did so when there was an icon.

hardinfo/info.c
bp0: I reverted that change.

includes/info.h
bp0: I've changed to booleans, although I was thinking of reasons to keep flags, like marking which fields (name, value, tag) should be freed.

lpereira: Yes, I've added another flag to determine if we should free the name field too -- used that to fix the Vulnerabilities thing.

Issue #348, opened 2019-06-18
Installing GTK+3 version on Clear Linux installs in "/usr/usr/..."


lpereira:
More information here: https://github.com/clearlinux/distribution/issues/270

PR #396, opened 2019-07-03
Update GNUInstallDirs.cmake

bp0:
Maybe it will help with #348. 
Doesn't seem to hurt anything using a newer version, but I haven't tried with very old cmake. Maybe @mckaygerhard could help out with that.


bp0:
Looks like this was already merged.

PR #398, opened 2019-07-04
shell: fix icons with flags [v3]

bp0:
See: #391
I split this out of #393 because it isn't really related to the struct Info stuff, it just fixes a bug in the existing icon handling.


bp0:
I've rebased with master and updated it to work with the new view type.

lpereira:
Merged, thanks!

PR #399, opened 2019-07-04
Add icons to memory devices

bp0:
Requires #398.


bp0:
I'll wait for #398 to be merged.

PR #400, opened 2019-07-05
devices/storage: current media fix

ocerman:
There was small mistake in my last PR: it could display incorrect current media in some cases. This is the fix.


lpereira:
Merged, thanks!

PR #405, opened 2019-07-06
computer/security: colored light icons for CPU vulnerabilities

bp0:
Image Source: https://www.flaticon.com/free-icon/traffic-light_196787
Image license: https://file000.flaticon.com/downloads/license/license.pdf
Attribution required, may alter and create derivative works.

Looks like:
vuln_lights


lpereira:
The problem in using only colors is that this makes it harder for color-blind people to distinguish those icons.  Using the :warning: or :heavy_check_mark: are better choices.  Even :+1: might be an issue in some cultures...

bp0:
Well, in this case the icons aren't by themselves. The text is right there next to it for the colorblind folks. 
Anyway, it's just an idea.

bp0:
Oh, I forgot to mention, there is a little demonstration in this PR where I used 128x128 sized icons and they work well. In the HTML view they will automatically get better with zoom or on HiDPI monitors, and since HardInfo is already using icon_cache_get_pixbuf_at_size() we could maybe adjust the size requested for that as well, based on the display.

bp0:
@lpereira, hows this? Changed to color+signs, and using SVG now, so it will need #407's SVG support in HTML reports patch.
vuln_lights2

bp0:
Now that the icons have shapes and are not color-only, do you still have an objection?

lpereira:
Icons look better with distinctive shapes, but it would be nicer if they were using the same style as the rest of the icons.

bp0:
I've changed the green and red to match colors from existing icons. The green is from memory.png and the red is from dialog-error.png. They look pretty good with the others, in my opinion. I don't know that I have the talent to try and make new graphics that match the style of the old ones, so I'd ask that you just accept this PR and then replace the images with ones that you would prefer. I really think that going with SVG is a good idea as the old PNGs will look very bad on high resolution (FHD+) displays. I also think that there is value in having a generic set of "good", "ok/mixed", and "bad" icons like these.

Here is a screen capture with the tweaked colors:
Screenshot_2019-11-23_00-57-21

mckaygerhard:
hi guys.. great to see how are growing hardinfo.. .. i have friend blind person, testing in right way you must use orca.. currently i'm very busy but i noted we need a nithly build system.. of statically linked binaries for those test.. .. i dont remenber the GTK API but you @bp0 must take care that GTK2 and GTK3 can complaint/integrate with atspi api correctly (there's no so much rather than use the GTK right funtions.. that have support, like pass text inclusivelly if have a icon assigned)

this it's not the only case @lpereira.. seems there's some code in old inherit that need some work too.. i'll try to compile a package and pass to my friend and report feedback

Issue #404, opened 2019-07-06
Show Ubuntu flavor


bp0:
There are several flavors of Ubuntu that all show as "Ubuntu" where the flavor could be reported instead. This could be done with:
```
apt-cache policy ubuntu-server ubuntu-desktop xubuntu-desktop kubuntu-desktop lubuntu-desktop ubuntu-mate-desktop ubuntu-budgie-desktop ubuntukylin-desktop ubuntustudio-desktop
```
Output like:
```
ubuntu-server:
  Installed: (none)
  Candidate: 1.431
  Version table:
     1.431 500
        500 http://us.archive.ubuntu.com/ubuntu disco/main amd64 Packages
ubuntu-desktop:
  Installed: (none)
  Candidate: 1.431
  Version table:
     1.431 500
        500 http://us.archive.ubuntu.com/ubuntu disco/main amd64 Packages
xubuntu-desktop:
  Installed: 2.228
  Candidate: 2.228
  Version table:
 *** 2.228 500
        500 http://us.archive.ubuntu.com/ubuntu disco/universe amd64 Packages
        100 /var/lib/dpkg/status
...
```

bp0:
Someone else will have to do Fedora spins, I don't use it anywhere.

PR #407, opened 2019-07-06
computer/os: Ubuntu flavors

bp0:
As I wished in #404: Show the specific Ubuntu flavor that is installed by checking for packages.

This also includes a new feature for struct Info, `info_field_skip()` macro, to allow conditional fields to be in the list sent to `info_add_group()`. @lpereira may want to implement it differently, I just abused the `.update_interval` member.


modules/computer.c
```diff
@@ -563,15 +563,19 @@ gchar *callback_os(void)
         info_field(_("C Library"), computer->os->libc),
         info_field(_("Distribution"), computer->os->distro,
                    .icon = distro_icon),
+        computer->os->ubuntu_flavor
```
lpereira: `info_group_add()` returns a group pointer, so it's possible to add these optional fields without `info_field_skip()`:

```C
InfoGroup *version = info_add_group(info, _("Version"), ...,
                                    info_group_last());
if (computer->os->ubuntu_flavor) {
    info_group_add_fieldsv(group,
                          info_field(_("Spin/Flavor"),
                          computer->os->ubuntu_flavor->name,
                          .icon = computer->os->ubuntu_flavor->icon),
                          info_field_last());
}
```

Feel free to add a `info_group_add_field()` function that adds a single field to a group -- so you don't need `info_field_last()` here.

I'd like to avoid adding magic numbers and flags at this point; the idea is to clean up the code, removing anything not obvious.  The strings in GKeyFile are really difficult to understand already.

bp0: Alright. New version uses info_group_add_field().

PR #409, opened 2019-07-07
devices/storage - added more partitions info

ocerman:
When using udisks2 >= 2.7.2, partitions list now shows partition label, size and filesystem


modules/devices/storage.c
```diff
@@ -24,11 +24,13 @@
 
 gchar *storage_icons = NULL;
 
+// udisks2 support by Ondrej Čerman
```
lpereira: You can add the credits to the file header next to the copyrights.

ocerman: Moved

PR #411, opened 2019-07-07
[discussion] struct info free/no-free sets

bp0:
@lpereira, what do you think about just having two sets for name, value, icon, tag, where one set is const and the other always gets freed? The the proper member can just be set for whichever is appropriate. If the string should be owned by the info field struct set .name, if not set .name_const, etc, This replaces the free_on_flatten members for each one. The info_field_get_*() macros return whichever is set for convenience. 
This PR implements the idea in info.{h,c} but doesn't change any uses of it, so don't expect it to compile.

### DIFF
--- a/hardinfo/info.c
+++ b/hardinfo/info.c
@@ -40,7 +40,7 @@ void info_group_add_fieldsv(struct InfoGroup *group, va_list ap)
     while (1) {
         struct InfoField field = va_arg(ap, struct InfoField);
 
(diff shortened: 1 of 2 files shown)


lpereira:
We can make everything non-const and just have a boolean to free -- a lot of places will copy the Info struct, and pointers are bigger; booleans are 1-byte only (barring alignment issues), so should be slightly better.

Issue #197, opened 2017-09-28
New shell


lpereira:
@bp0, you've mentioned on another issue something about working on a new shell. What do you have in mind?

I've been thinking about changing the UI a little bit, so it's closer to what HardInfo was in previous releases:

![0.3](http://web.archive.org/web/20050123065637im_/http://alpha.linuxmag.com.br:80/~leandro/hardinfo/hardinfo-ptbr.png)

The idea is, when using GTK+3, to use the new header bar, like the GNOME System Monitor, to switch between modules, and have a list on the left with each module entry:

![systemmonitor gtk3](https://i.imgur.com/g2vA9QL.png)

The GTK+2 version can use tabs:

![systemmonitor gtk2](http://www.pedroliveira.pt/wp-content/uploads/2015/08/gnome-system-monitor.png)

bp0:
My idea was to have three tabs in the GTK header bar. 
* A tab with the detailed system information, mostly as it is now.
* A tab like System Monitor above, but using uber-graph, with selectable items in the graph from any of the information. Instead of only in having the graph in Memory or Sensors. 
* A tab for the Benchmark functions and results.
I'd like to have the monitor tab continue to update during the benchmark, tracking whatever you've selected to monitor. CPU or thermal sensor, or whatever.
I've only started playing with it, but I've had no free time lately to continue.

Since uber only works with gtk3 anyway, I built a separate shell3.c for the new gtk3-only version. This does away with the need for all the gtk version conditional code. I think the gtk2 version could stay as it is.

mrnhmath:
There are some GTK3 DEs that doesn't use header bars, like MATE, Cinnamon and the next version of Xfce. An option to retain the title bar + menu bar + toolbar model would be appreciated.

mckaygerhard:
@mrnhmath lxde will never drop support for GTK2 as started example, that's why GTK2 are still in hardinfo, very probed and knowed!

to drop completely GTK2 important project must drop, so GTK2 them not used but as can see, GTK2 still areused in many projects due performance respect gtk2 and the GTK3 api its a desaster (that caues LXDE started the LXQT proyect)

all enought reasons good explained

Issue #180, opened 2017-08-20
struct Info


bp0:
Some ideas about struct Info.

## All strings version of info_add_group(), info_add_group_ff()
Could make things tidier, or not. See #179.

## Perhaps the label and value should be cleaned
Implies, like info_add_group_ff() above, all values become owned by this struct and freed. Would need to signal if HTML is actually being used in the value, and then cleaning would be skipped. See info_field_full() below.
```
[Snip from hardinfo.h]
/* Hardinfo labels that have # are truncated and/or hidden.
 * Labels can't have $ because that is the delimiter in
 * moreinfo.
 * replacing = true will free v */
gchar *hardinfo_clean_label(const gchar *v, int replacing);
/* hardinfo uses the values as {ht,x}ml, apparently */
gchar *hardinfo_clean_value(const gchar *v, int replacing);
```

## Mark for default selection
Implement the mark for default selection feature from 5b12d5aca5fbebfc3656c5abd903532f9ed1282c

## Extended version of info_field(), info_field_full()
Something like
```
struct InfoField info_field_full(
    const gchar *tag, /* equiv of old "$tag$label=value\n" format */
    const gchar *name,
    const gchar *value, 
    gboolean select, /* mark to select */
    gboolean html, /* don't clean */
    ...
);
```
## Handle moreinfo* in this struct instead of separately 
It's already very awkward to handle these separately.

## Replace hi_get_field() with a get_value() function that can use a function pointer member of struct InfoField.
I think the InfoField should be able to specify a function that should be called to update the value.  Also, hi_get_field() uses translated label names to update fields. If the tag that is used currently used by moreinfo is available, maybe let's use that.

## Replace hi_exported_methods() by using the untranslated tag to get a value from a struct InfoField
This has a few implications that I'm not sure about yet.

lpereira:
Another thing that I'd like to see in Info is the ability to create MoreInfo.  For instance, the firmware view could benefit from this right now.

The API can be simple, I think.  Something like `struct Info *info_add_item(struct Info *root, struct InfoItem item)`, with a macro similar to `info_field()` so we can have optional parameters and whatnot.  This will add a new Info to a list in a struct Info*, and when flattening, we know that if that list isn't empty, we should use moreinfo.  Since it returns a normal `struct Info`, we can use groups and fields as usual.  It should return NULL when adding a third level because the shell at the moment can't handle this.

PR #416, opened 2019-07-12
struct info unflatten [v2]

bp0:
I started this a while ago (2017, I think, see #180 and #197) when I was trying to make a new shell that only operated using struct Info. The info_unflatten() function makes an Info struct from the old shell key/value file format string. Trying to do it exposed many problems with the Info struct at the time and I gave up trying to solve them all, and gave up on the new shell.

I've updated the function for more recent changes to the Info struct and shell, but there are still a few problems. For example, group.name always leaks from an unflatten()-ed Info.

This function seems very useful for the original purpose: shell could be rewritten to be Info-only without changing everything else. I've also found it can be useful for debugging.

### DIFF
--- a/hardinfo/info.c
+++ b/hardinfo/info.c
@@ -355,3 +355,128 @@ gchar *info_flatten(struct Info *info)
 
     return g_string_free(values, FALSE);
 }
+
+#define SEQ(a,b) (g_strcmp0(a,b) == 0)
+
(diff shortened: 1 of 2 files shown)


lpereira:
Seems to be a good strategy to attack the work to change the shell over to
Info structs.  Will look with care, as I still have very spotty internet
access.

PR #417, opened 2019-07-13
vendor.ids update and vendor.{h,c} improvements from sysobj version

bp0:
* more fields in Vendor struct.
* match_case -> match_rule and new match rule "exact". match_rule
  remains compatible with match_case for old conf format.
* matches must be "whole word" matches so "Harmony" no longer hits
  for "ARM".
* Parts outside of () are checked first, so "Foo (formerly Barly)"
  matches Foo before Barly, even though Barly is longer and would
  otherwise match first.
* vendor_list type (a blessed GSList*) and helper functions to manage
  it.

### DIFF
--- a/deps/sysobj_early/gui/uri_handler.c
+++ b/deps/sysobj_early/gui/uri_handler.c
@@ -1,0 +1,48 @@
+/*
+ * sysobj - https://github.com/bp0/verbose-spork
+ * Copyright (C) 2018  Burt P. <pburt0@gmail.com>
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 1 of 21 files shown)


bp0:
I also noticed that the new detail view link handling didn't always work right, so I've included the uri_handler functions as well to fix that.

deps/sysobj_early/gui/uri_handler.c
```diff
@@ -0,0 +1,48 @@
...
+    gboolean ret = FALSE;
+    if (uri_func)
+        ret = uri_func(uri);
+    if (ret) return TRUE;
+
+    return uri_open_default(uri);
+}
+
+gboolean uri_open_default(const gchar *uri) {
+    gchar *argv[] = { "/usr/bin/xdg-open", (gchar*)uri, NULL };
```
lpereira: HardInfo already has `url_open()`, which will try to use `xdg-open` if available (with a few more fallbacks -- some of which might not even exist any longer).  Might be worth using it instead (or maybe even copying it to `verbose-spork`).

bp0: I looked at the open_url() function in hardinfo/util.c when I made the uri_handler, and decided on the simplified version. xdg-open has been standard, as far as I can tell, for ~18 years, so I figured that I could count on it existing anywhere that GTK2.0+ does (only 17 years ago).  I think maybe replace the few open_url() with uri_open() and removing it might be better.

bp0:
@ocerman,  Please review 8cc11c8678babb05ccb63d2ed2001979e2a11dd9, which modifies udisks2_util to add vendor information for SD/MMC cards.

deps/sysobj_early/gui/uri_handler.c
lpereira: Yeah, you're right.  That code is indeed pretty old.  No worries about using `xdg-open` directly and junking that thing.

Issue #379, opened 2019-06-30
Hyperlinks for vendor websites


lpereira:
Wrote this nasty hack here, that might be useful to have vendor links in HardInfo: https://gist.github.com/lpereira/4178c6b4bc6f2de9b4ad02b712d06389

Only allows one link per row, and can probably be cleaned up a lot...

The only missing piece right now is how to teach the report generator and the shell what a link is.  Maybe use something neutral like markdown syntax?

lpereira:
I remembered there was a markdown parser in HardInfo a few years ago, part of the (scrapped) help viewer.  It's able to output text, pango, and HTML as well. Might be just the ticket.

lpereira:
Spent some time on this. This will happen, but not using that nasty hack; decided to add a new ShellViewType that renders things using labels and tables instead. Labels can have links.

bp0:
"Copy to Clipboard" command copies with markup (Pango's HTML subset). Maybe you could fix that too while working on this bit?

bp0:
Also note that labels may only have links in GTK>2.18 so the requirement needs to be raised.
This was talked about before but there was an objection. See #118.

bp0:
An alternate thing would be to use a hinote-like extra pane with all the vendor information in it when an item with vendor information is selected similar to what I did here:
![vendors_as_box](https://user-images.githubusercontent.com/1758090/60549293-0c0d7680-9cea-11e9-892c-4984793bd901.png)

lpereira:
Here are some screenshots showing how it's going to look.

This has been a long time coming, and should fix a bunch of stuff:
- The Summary view will eventually use this thing here.  Summaries can't be included in reports, can't be copied to the clipboard, and fields can't be updated automatically.
- This is currently a new ViewType. Probably will stay that way, but I also want to have the same view for the MoreInfo panel.
- It doesn't currently support icons, updating fields automatically, etc.  Will continue working on this.
- Supporting older GTK+ versions is possible; my idea is to use the markdown parser that I used to use in the [help-viewer](https://github.com/lpereira/help-viewer/), which was included in HardInfo at some point. It's able to output plain text, HTML, or Pango markup, which is great for us, as we need all of those.
- I'll also take a closer look at the GNOME HIG to make this with the expected spacing and whatnot.
- Things like the Input Devices in the Computer->Summary screen will actually be listed as a bunch of bullet points.

[...]

PR #418, opened 2019-07-13
Raise GTK2 requirement from 2.10 to 2.18

bp0:
As links in labels are now being used, 2.18 is required.

See: https://github.com/lpereira/hardinfo/issues/379#issuecomment-507857291

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,7 +81,7 @@ if (HARDINFO_GTK3)
 	add_definitions(-DGTK_DISABLE_SINGLE_INCLUDES)
 else()
 	message(STATUS "Building for GTK2")
-	pkg_check_modules(GTK REQUIRED gtk+-2.0>=2.10 glib-2.0>=2.10 gthread-2.0>=2.10 gmodule-export-2.0>=2.10)
+	pkg_check_modules(GTK REQUIRED gtk+-2.0>=2.18 glib-2.0>=2.10 gthread-2.0>=2.10 gmodule-export-2.0>=2.10)
 endif()
 
 if(NOT HARDINFO_NOSYNC)
--- a/shell/callbacks.c
+++ b/shell/callbacks.c
@@ -115,11 +115,7 @@ void cb_about_module(GtkAction * action)
 	    gtk_window_set_transient_for(GTK_WINDOW(about), GTK_WINDOW(shell->window));
 
 	    text = g_strdup(sm->name);
-#if GTK_CHECK_VERSION(2, 12, 0)
 	    gtk_about_dialog_set_program_name(GTK_ABOUT_DIALOG(about), text);
-#else
-	    gtk_about_dialog_set_name(GTK_ABOUT_DIALOG(about), text);
-#endif
 	    g_free(text);
 
 	    gtk_about_dialog_set_version(GTK_ABOUT_DIALOG(about),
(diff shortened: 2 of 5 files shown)


mckaygerhard:
hi! i see the comment and well, today.. a recent version of hardinfo only compiles in recent version of gcc and linux.. that again does are not well performed in those arm based old devices.. 

i was the objection in that past comment.. but i cannot able to compile hardinfo in those rasberry devices.. and i have many.. i Debian 10, fedora 7 are too heavy distributions for that falvors devices.. so compile recent hardinfo version are now nonsense.. 

we used hwinfo instead.. **the cause: issue  related to pull #160 since commit 3f18100  so if hardinfo does not compiled in glibc << 2.11 there's no sense for those older systems still running kernel 2.6 and 3.0, due unsuported hardware in recent kernels like good pool network cameras and security related, also many spensive multimedia devices wtill good working but very older**

bp0:
After this change, requires: cmake 2.6+, gcc 4.6+, gtk 2.18+, glib 2.10+. So any of Debian 5.0+, Ubuntu 8.04+, Slackware 13.0+, or about 10 years... if you can get the source downloaded on such old systems, as git clone will not work, neither will wget be able to connect with SSL to github.

As for hardware I've tested on Raspberry PI 1B, 1x ARM11 900Mhz, Raspbian 8 Jesse, and on an old 1.4Ghz Pentium 4, and all worked.

I really like to support old Linux and old hardware as much as possible, but also like new features. It's unfortunate that hardinfo doesn't have a stronger separation between the core and the GUI bits so that they could have different requirements.

mckaygerhard:
> unfortunate that hardinfo doesn't have a stronger separation between the core and the GUI bits so that they could have different requirements.

well consider that hardinfo are more than 10+ years old .... in those days the client-server behaviour was (and still are) not the most faster..  a socket/client separation are themost good behaviour in unix-linux like environments

> After this change, requires: cmake 2.6+, gcc 4.6+, gtk 2.18+, glib 2.10+. So any of Debian 5.0+, Ubuntu 8.04+, Slackware 13.0+, or about 10 years... 

that raised to any Debian 7.0+, Winbuntu 14+, Slack 13+

> if you can get the source downloaded on such old systems, as git clone will not work, neither will wget be able to connect with SSL to github.

my wget backport works.. also my git under Debian ETCH !!!

bp0:
I tried hardinfo in void linux, which has no systemd, and no udev for udisks2, and it worked. The fallback methods for storage include lsscsi, which does actually report nvme devices, and /proc/scsi/scsi which does not report nvme, but it only available on older kernels which probably don't support nvme anyway? Do you have a Debian etch system with an nvme device?

I tried to test etch, but it is not supported by debian anymore. The archive is not available, and the DVDs can't even be downloaded anymore. I'm curious what exact board you might have that <s>is</s> <b>WAS</b> supported by Debian etch, but not by void linux currently?

lpereira:
Maybe for the links we can `#ifdef` and not generate one, so that we can release a last version supporting older GTK+ versions?  Or is that not the case anymore, and 2.18 is indeed the minimum supported version?

bp0:
Ok lets just put in #ifdefs, but honestly does anyone have an idea what old version of GTK is really supported? As I mentioned, GTK 2.18 was added in Debian 5 Lenny (February 2009). I tried to build hardinfo in earlier debian, but failed because git and openssl were too old to even fetch the source, and I didn't care to continue.

bp0:
Close in favor of #436.

PR #423, opened 2019-07-21
shell/info: allow newlines in values

bp0:
Also fixed a FIXME: use g_key_file_get_string_list().


lpereira:
Why do you need `\n`?  Isn't it better to create more items?  I'm actually planning to add support for bullet lists in the Info struct (generating keys with `•#some_number`), and use it for other things (e.g. printer features, udisks2 features, etc.)

bp0:
I think even if it isn't used, it is better this way than before.

PR #427, opened 2019-07-26
Improved USB devices

ocerman:
This PR adds sysfs information about USB devices.
This brings additional information (speed, used driver, interface label, reported manufacturer and device name) and also sysfs can be used as a fallback when lsusb is not available.

This PR also contains workaround when same USB interface is parsed multiple times from lsusb output.


bp0:
lgtm

PR #434, opened 2019-08-09
usb: use sysfs info by default

ocerman:
This PR:
 - adds usb.ids parsing (from @bp0 sysobj), so now it is possible to get all usb info without lsusb.
 - added more properties like serial number
 - because lsusb is slower than parsing stuff from sysfs, sysfs method is now used by default (lsusb is called only when sysfs method fails).

### DIFF
--- a/hardinfo/usb_util.c
+++ b/hardinfo/usb_util.c
@@ -64,8 +64,10 @@ void usbd_free(usbd *s) {
         g_free(s->device);
         g_free(s->usb_version);
         g_free(s->device_version);
+        g_free(s->serial);
         g_free(s->dev_class_str);
         g_free(s->dev_subclass_str);
+        g_free(s->dev_protocol_str);
         g_free(s);
     }
 }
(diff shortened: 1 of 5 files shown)


hardinfo/usb_util.c
```diff
@@ -279,17 +298,53 @@ static gboolean usb_get_interface_sysfs(int conf, int number,
...
+void find_usb_ids_file() {
+    if (usb_ids_file) return;
+    char *file_search_order[] = {
+        g_build_filename(g_get_user_config_dir(), "hardinfo", "usb.ids", NULL),
+        g_build_filename(params.path_data, "usb.ids", NULL),
+        NULL
+    };
+    int n;
+    for(n = 0; file_search_order[n]; n++) {
+        if (!access(file_search_order[n], R_OK))
```
bp0: There's a bug. Should be `if (!usb_ids_file && !access(file_search_order[n], R_OK))`
The loop needs to complete to free the unused ones, but it should only be set for the first one found.
My fault, I'll have to fix it for arm.ids and sdcard.ids as well.

ocerman: np, I pushed the fix

PR #441, opened 2019-08-11
benchmark cleanup and fixes, add sysbench

bp0:
Screenshot_2019-08-13_13-39-38

## update nice_name_x86_cpuid_model_string()
  - `remove_long_core_count`: Removes "Quad Core", "Six-Core", etc., since the CPU config is shown already next to the name. Marketing names like "Core 2 Quad" and "Athlon X2" remain.
  - `remove_amd_compute_cores`: now turned on for the same reason as `remove_long_core_count`

## benchmark.conf 
  - passed CPUs (new results only) through the nice_name_x86_cpuid_model_string() function to clean them up
  - fixed old results with smooshed names, so that they would be cleaned properly when loaded. The old results can not be cleaned because they still need to be unique keys.
  - removed empty result blocks for non-existing benchmarks
  - ~removed old blowfish result block~

## Other changes 
* add wrapper for sysbench, memory benchmark
* fix zlib display order
* fix cryptohash MiB/s calculation -- changes results slightly
* revision and params for other benchmarks
* allow attaching user note to bench result with -u
* don't inlcude the new result value bits if they are empty/invalid
  in bench_value_to_str(). bench_value_from_str() doesn't need to be modified.
* bench_results: clean old result cpu name for x86
* use problem_marker() from dmi_memory to mark old version bench results
* escape strings in bench result conf lines
* add machine_data_version field
* benchmark: verifiable test data size and content
  - The test data benchmark.data is stored in a file
    that could be edited to change the size or content.
    /* to guarantee size */
    gchar *get_test_data(gsize min_size);

    /* to checksum content */
    char *md5_digest_str(const char *data, unsigned int len);

## just_bench.sh

Until the whole bench submission framework is complete,
here is the best way to submit bench results to hardinfo.

* Run `test/just_bench.sh` to create a conf with results
* Merge those results into benchmark.conf using crudini
* Submit a pull request

### example:
```
bash test/just_bench.sh >whatever.conf
crudini --merge data/benchmark.conf <whatever.conf
```
... commit, push, pull request.


lpereira:
Lots of good stuff here, thanks!

PR #443, opened 2019-08-11
vendor: vendor_get_link_from_vendor() use wikipedia in some cases

bp0:
Some vendors, like those defunct, may not have url, but may have
a wikipedia article. Use wikipedia in the link in those cases.

Also, Wikipedia offers more information than the vendor's site
in most cases, and has the vendor's url right there in the sidebar
to click, if desired. I think using wikipedia for links is
prefereable, but this could be turned off by setting
link_wikipedia to FALSE.


lpereira:
Good idea

PR #446, opened 2019-08-14
 report: show icons for module entry in HTML report 

### LINKED ISSUE #444: Add module entry icons to HTML reports
Shouldn't be too hard, eh?

bp0:
See #444


lpereira:
Nice!

PR #449, opened 2019-08-17
Devices/Monitors

bp0:
Adds a new entry in Devices that lists the attached monitors with their EDID data.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -295,7 +295,7 @@ add_executable(hardinfo
 	shell/loadgraph-uber.c
 )
 target_link_libraries(hardinfo
-	sysobj_early
+	-Wl,-whole-archive sysobj_early -Wl,-no-whole-archive
 	uber-graph
 	${GTK_LIBRARIES}
 	${LIBSOUP_LIBRARIES}
@@ -333,7 +333,7 @@ add_executable(hardinfo
 	shell/loadgraph.c
 )
 target_link_libraries(hardinfo
-	sysobj_early
+	-Wl,-whole-archive sysobj_early -Wl,-no-whole-archive
 	${GTK_LIBRARIES}
 	${LIBSOUP_LIBRARIES}
 	m
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,7 @@ set(HARDINFO_RESOURCES
 	"deps/sysobj_early/data/sdcard.ids"
 	"deps/sysobj_early/data/usb.ids"
 	"deps/sysobj_early/data/arm.ids"
+	"deps/sysobj_early/data/edid.ids"
 )
 
 set(HARDINFO_MANPAGES
@@ -152,6 +153,7 @@ set(MODULE_devices_SOURCES
 	modules/devices.c
 	modules/devices/${HARDINFO_ARCH}/processor.c
 	modules/devices/gpu.c
+	modules/devices/monitors.c
 	modules/devices/battery.c
 	modules/devices/dmi.c
 	modules/devices/dmi_memory.c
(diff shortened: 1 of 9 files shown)


lpereira:
With this patch, you can get rid of `wtf()`:

```patch
diff --git a/hardinfo/x_util.c b/hardinfo/x_util.c
index 07d5226..1508f52 100644
--- a/hardinfo/x_util.c
+++ b/hardinfo/x_util.c
@@ -162,9 +162,9 @@ gboolean fill_xrr_info(xrr_info *xrr) {
 
     memset(&ts, 0, sizeof(x_screen));
     memset(&to, 0, sizeof(x_output));
-    memset(output_id, 0, 128);
-    memset(status, 0, 128);
-    memset(alist, 0, 128);
+    memset(output_id, 0, sizeof(output_id));
+    memset(status, 0, sizeof(status));
+    memset(alist, 0, sizeof(alist));
 
     spawned = hardinfo_spawn_command_line_sync(xrr_cmd,
             &out, &err, NULL, NULL);
@@ -195,7 +195,7 @@ gboolean fill_xrr_info(xrr_info *xrr) {
             /* looking for:
              * <output_id> (connected|disconnected|unknown connection) (primary|?) <%dx%d+%d+%d> (attribute_list) mm x mm
              */
-            ec = sscanf(p, "%s %[^(](%[^)]", output_id, status, alist);
+            ec = sscanf(p, "%127s %127[^(](%511[^)]", output_id, status, alist);
             if (ec == 3) {
                 int is_output = 0, found_rect = 0, n = 0;
                 gchar **ot = g_strsplit(status, " ", 0);
```

bp0:
It's a nice patch, but I still get the crash.

For me, Hardinfo crashes on selecting "Monitors" with the message:
```
hardinfo: symbol lookup error: /usr/local/lib/hardinfo/modules/devices.so: undefined symbol: edid_new
```
But `objdump -t libsysobj_early.a`:
```
...
0000000000000bd0 g     F .text	0000000000003152 edid_new
...
```

I was stumped, but I discovered that if I call the function from somewhere else, then it would not crash, so I added that wtf() function there to get by this frustrating block. The strangest thing is that wtf() is never even called or referenced anywhere, but having code chunk there prevented the crash somehow.

I'd be really happy to know what is going on there.

lpereira:
It's a different kind of crash; without that patch, there's a stack overflow in `fill_xrr_info()` right after HardInfo starts.  I'm also getting this crash after I select the Monitors entry.  No symbols containing `edid` in their name appear in the hardinfo binary, so they're probably being stripped out because they're in a static library.  One thing you might try doing is, instead of linking with `sysobj_early`, link with `-Wl,-whole-archive sysobj_early -Wl,-no-whole-archive` instead.

bp0:
I'm not sure I know how that would be done with cmake. 
I wonder why it would be stripped out of hardinfo anyway? and why isn't it stripped out if some other part of sysobj_early references it?

lpereira:
Just change the line in `target_link_libraries()`. It should work. (I do that in Lwan: [here](https://github.com/lpereira/lwan/blob/master/CMakeLists.txt#L250-L254) and [here](https://github.com/lpereira/lwan/blob/master/src/bin/lwan/CMakeLists.txt#L6).)

This happens if you're linking with a static library.  Even though they're just an archive with .o files, the behavior is different than linking with the actual .o files that are inside the archive. I don't know why the behavior is different, though.

bp0:
Thank you, it worked. I'm only a little bothered that I don't know why. It will pass.

lpereira:
Found it in the `ld` man page:

```
       --whole-archive
           For each archive mentioned on the command line after the --whole-archive option,
           include every object file in the archive in the link, rather than searching the
           archive for the required object files.  This is normally used to turn an archive file
           into a shared library, forcing every object to be included in the resulting shared
           library.  This option may be used more than once.
```

lpereira:
Merged, thanks!

Here's the output on my laptop (the `12.5""` looks weird, maybe that's an escaping bug?):

Screenshot from 2019-08-18 14-49-01

PR #461, opened 2019-08-31
Update pt_BR 31082019

pepaulo:


bp0:
There should not be a hardinfo.po, but instead the hardinfo.pot should be updated by running `updatepo.sh` in the `po/` directory.

PR #473, opened 2019-10-23
devices/storage: Added normalized values of reported S.M.A.R.T. attributes

ocerman:
smart


lpereira:
LGTM, even though I'd prefer to see these values in a proper table (which is out of scope for this PR, of course).

PR #474, opened 2019-10-24
Fix FR build problems

bp0:
make fails after 8fd37c17522681918345c0d0f4b8cd561f4f74a6 due to errors in fr.po. First commit removes the problem strings, second commit re-syncs the strings from hardinfo.pot because a couple of problems were translations of older versions of c-strings.


lpereira:
Thanks!

PR #477, opened 2019-10-29
x86/processor: bigger buffer for cpuinfo

ocerman:
Removes warning `Warning: truncated a line (probably flags list) longer than 896 bytes while reading /proc/cpuinfo.`

### DIFF
--- a/modules/devices/x86/processor.c
+++ b/modules/devices/x86/processor.c
@@ -454,8 +454,9 @@ GSList *processor_scan(void)
     GSList *procs = NULL, *l = NULL;
     Processor *processor = NULL;
     FILE *cpuinfo;
-    gchar buffer[PROC_SCAN_READ_BUFFER_SIZE];
+    gchar *buffer;
 
+    buffer = (gchar *) g_malloc(PROC_SCAN_READ_BUFFER_SIZE * sizeof(gchar));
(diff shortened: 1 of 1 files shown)


ocerman:
@bp0 true, I have pushed another commit with malloc-ed buffer.

bp0:
lgtm

modules/devices/x86/processor.c
```diff
@@ -448,14 +448,15 @@ gchar *caches_summary(GSList * processors)
...
+#define PROC_SCAN_READ_BUFFER_SIZE 1024
 GSList *processor_scan(void)
 {
     GSList *procs = NULL, *l = NULL;
     Processor *processor = NULL;
     FILE *cpuinfo;
-    gchar buffer[PROC_SCAN_READ_BUFFER_SIZE];
+    gchar *buffer;
 
+    buffer = (gchar *) g_malloc(PROC_SCAN_READ_BUFFER_SIZE * sizeof(gchar));
```
lpereira:
```suggestion
    buffer = g_malloc(PROC_SCAN_READ_BUFFER_SIZE);
```

(`sizeof(gchar)` is always going to be 1, and `gpointers` will always cast to any pointer automatically.)

lpereira:
Merged, thanks!

PR #479, opened 2019-11-09
memory devices: sum online blocks as fallback physical memory count

bp0:
if no eeprom for SPD, and no root for DMI, try online block count x block size.

### DIFF
--- a/modules/devices/dmi_memory.c
+++ b/modules/devices/dmi_memory.c
@@ -386,7 +386,6 @@ static void dmi_fill_from_spd(dmi_mem_socket *s) {
 }
 
 static dmi_mem_size size_of_online_memory_blocks() {
-    gchar **online_files = NULL;
     gchar *block_size_bytes_str = NULL;
     dmi_mem_size block_size_bytes = 0;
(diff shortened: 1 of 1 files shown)


modules/devices/dmi_memory.c
```diff
@@ -382,6 +385,34 @@ static void dmi_fill_from_spd(dmi_mem_socket *s) {
...
+    dmi_mem_size block_size_bytes = 0;
+    dmi_mem_size ret = 0;
+
+    if (g_file_get_contents("/sys/devices/system/memory/block_size_bytes", &block_size_bytes_str, NULL, NULL) ) {
+        block_size_bytes = strtoll(block_size_bytes_str, NULL, 16);
+    }
+    if (!block_size_bytes)
+        return 0;
+
+    online_files = get_output_lines("find /sys/devices/system/memory -name online");
```
lpereira: I would prefer not to call any shell command unless it's absolutely necessary.

bp0: Ok

lpereira:
Merged, thanks!

Issue #493, opened 2019-12-27
Memory Vendors string issue


ocerman:
![hardinfo-memi](https://user-images.githubusercontent.com/10187350/71522743-3ec5af80-28c6-11ea-854d-0cbaebb57175.png)

bp0:
I'd like to show the actual value read and the vendor match information. The way it is done now, aside from the missing space, looks redundant. I have been working on a solution where the actual value is used as the field value, with the vendor match info added below it:

![Screenshot_2019-12-27_18-24-08](https://user-images.githubusercontent.com/1758090/71536539-2067b000-28d6-11ea-8021-a4fea78d56da.png)
![Screenshot_2019-12-27_18-12-36](https://user-images.githubusercontent.com/1758090/71536540-2067b000-28d6-11ea-8a76-97e78ab71b39.png)

bp0:
Fixed in #494

PR #494, opened 2019-12-28
Vendor Info and colored vendor tags

bp0:
As I said in #493, I think it's better to have the vendor values be the actual data from the hardware or lookup, and the vendor.c match data along side it instead of making a link.

Reasons:
* The actual value is interesting by itself
* There's usually more info available for a vendor than one link
* There might be a mismatch

So instead of doing the vendor lookup ourselves, we can just mark the value as a (possible) vendor
value and then it will be looked up by the shell, and a little info box will be added below it for any match.

To mark the field value as a possible vendor string:
* In the old shell conf style use the flag `^`.
* struct Info style use `(struct InfoField).value_has_vendor`.
* shell function: `key_value_has_vendor_string()` to check for the flag

I've converted these: dmi, x86 processor, gpus, monitors, pci, usb, memory devices, storage, printers.

Also, you might notice that I brought the little colored vendor tags over from sysobj.

Here are some screenshots:
Screenshot_2019-12-27_19-59-14
Screenshot_2019-12-27_18-12-36
Screenshot_2019-12-27_18-24-08
Screenshot_2019-12-27_18-58-42
Screenshot_2019-12-27_19-52-14
Screenshot_2019-12-27_19-52-49

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -236,6 +236,7 @@ add_library(sysobj_early STATIC
 	deps/sysobj_early/src/nice_name.c
 	deps/sysobj_early/gui/uri_handler.c
 	deps/sysobj_early/src/util_edid.c
+	deps/sysobj_early/src/format_early.c
 )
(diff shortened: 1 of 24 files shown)


lpereira:
This looks really nice!

lpereira:
Merged, thanks!

Issue #495, opened 2019-12-28
Use PCI vendor for NVME storage


bp0:
Since no vendor is supplied for NVME, use the PCI vendor for the device. I know this is going to be the vendor of the controller and not the actual part, but it would be better than nothing, right?

Maybe @ocerman could do this?

pci vendor is at `/sys/block/<block_dev>/device/device/vendor`

bp0:
@ocerman , nevermind I did it in #496.

PR #496, opened 2019-12-28
Try and get vendor for NVMe from PCI

bp0:
As in #495

### DIFF
--- a/hardinfo/pci_util.c
+++ b/hardinfo/pci_util.c
@@ -20,6 +20,80 @@
 
 #include "hardinfo.h"
 #include "pci_util.h"
+#include "util_ids.h"
+
+gchar *pci_ids_file = NULL;
+
+void find_pci_ids_file() {
+    if (pci_ids_file) return;
+    char *file_search_order[] = {
+        g_strdup("/usr/share/hwdata/pci.ids"),
+        g_build_filename(g_get_user_config_dir(), "hardinfo", "pci.ids", NULL),
+        g_build_filename(params.path_data, "pci.ids", NULL),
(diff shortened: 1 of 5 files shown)


hardinfo/pci_util.c
```diff
@@ -20,6 +20,80 @@
...
+    gchar *ret = NULL;
+
+    ids_query_result result = {};
+    gchar *qpath;
+
+    if (!pci_ids_file)
+        find_pci_ids_file();
+
+    qpath = g_strdup_printf("%04x", id);
+    scan_ids_file(pci_ids_file, qpath, &result, -1);
```
lpereira: `pci_ids_file` can be NULL here if `find_pci_ids_file()` fails to find in all of the 3 locations. Might be a good idea to `if (!pci_ids_file) return NULL;` after calling `find_pci_ids_file()`.

hardinfo/pci_util.c
```diff
@@ -20,6 +20,80 @@
 
 #include "hardinfo.h"
 #include "pci_util.h"
+#include "util_ids.h"
+
+gchar *pci_ids_file = NULL;
+
+void find_pci_ids_file() {
```
lpereira: Unexported functions should be static.

hardinfo/pci_util.c
bp0: Yeah, but scan_ids_file() will handle it alright. It would just do some useless work making the request.

hardinfo/pci_util.c
bp0: Ok

Issue #497, opened 2019-12-28
bench_value_from_str() doesn't handle weird euro floats with comma for decimal


bp0:
Thanks to @ocerman 's bench result (4c386e89eb9bf28e43164b958374bee113b4cee7) it is discovered that sscanf("%lf") is unreliable!

Complete a benchmark in 0 seconds! The new Ryzen's are truly amazing.

ocerman:
I just noticed I have different problem - Look at these old i5s
![bench](https://user-images.githubusercontent.com/10187350/71542805-1f349280-296b-11ea-9a54-3fed887e7065.png)

PR #498, opened 2019-12-28
fix bench_value_from_str() for floats with , for decimal

bp0:
Fixes https://github.com/lpereira/hardinfo/issues/497

### DIFF
--- a/modules/benchmark.c
+++ b/modules/benchmark.c
@@ -69,8 +69,8 @@ bench_value bench_value_from_str(const char* str) {
         if (c >= 3) {
             if ((p = strchr(rstr, ','))) { *p = '.'; }
             if ((p = strchr(estr, ','))) { *p = '.'; }
-            ret.result = strtod(rstr, NULL);
-            ret.elapsed_time = strtod(estr, NULL);
+            ret.result = g_ascii_strtod(rstr, NULL);
+            ret.elapsed_time = g_ascii_strtod(estr, NULL);
             ret.threads_used = t;
         }
         if (c >= 4) {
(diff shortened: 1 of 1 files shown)


ocerman:
This does not seems to work with my locale:
bench2

Seems like `strtod` is using current locale and because "," is correct decimal separator on my system, now even my result does not display correctly.

ocerman:
@bp0 
Maybe it would be better to save benchmark results in locale-indepedent format and then parsing it would be easier.

ocerman:
OK, I found solution: using `g_ascii_strtod` instead of `strtod` works fine.

modules/benchmark.c
```diff
@@ -60,14 +60,17 @@ char *bench_value_to_str(bench_value r) {
...
     if (str) {
-        c = sscanf(str, "%lf; %lf; %d; %d; %255[^\r\n;|]; %255[^\r\n;|]", &r, &e, &t, &v, extra, user_note);
+        /* try to handle floats from locales that use ',' or '.' as decimal sep */
+        c = sscanf(str, "%[-+0-9.,]; %[-+0-9.,]; %d; %d; %255[^\r\n;|]; %255[^\r\n;|]", rstr, estr, &t, &v, extra, user_note);
         if (c >= 3) {
-            ret.result = r;
-            ret.elapsed_time = e;
+            if ((p = strchr(rstr, ','))) { *p = '.'; }
+            if ((p = strchr(estr, ','))) { *p = '.'; }
+            ret.result = strtod(rstr, NULL);
```
ocerman: `g_ascii_strtod` is local-independent and `.` is always decimal separator

modules/benchmark.c
```diff
@@ -60,14 +60,17 @@ char *bench_value_to_str(bench_value r) {
...
+            ret.elapsed_time = strtod(estr, NULL);
```
ocerman: same as above

bp0:
Right on. 

Maybe integer of fraction of seconds maybe better, idk.

lpereira:
Heh. "eurofloat".

PR #500, opened 2019-12-28
Updated input devices

ocerman:
- Added vendor / product name from usb.ids
- Added new type "audio" for ALSA devices
- Added columns Vendor tag and Type
- Fixed computer/summary output


lpereira:
Merged, thanks!

PR #508, opened 2020-01-01
[WIP] storage: updated device label

### LINKED ISSUE #507: NVMe SSD vendor name doubled
double-samsung

ocerman:
FIX #507 

@bp0 can you check whether this works well also for your SSD?

### DIFF
--- a/hardinfo/info.c
+++ b/hardinfo/info.c
@@ -109,7 +109,7 @@ struct InfoField info_field_printf(const gchar *name, const gchar *format, ...)
     };
 }
 
-static void info_group_strip_extra(struct InfoGroup *group)
+void info_group_strip_extra(struct InfoGroup *group)
 {
     guint fi;
     char *val, *oldval;
--- a/includes/info.h
+++ b/includes/info.h
@@ -76,6 +76,7 @@ struct InfoField {
 struct Info *info_new(void);
 
 struct InfoGroup *info_add_group(struct Info *info, const gchar *group_name, ...);
+void info_group_strip_extra(struct InfoGroup *group);
 
 void info_add_computed_group(struct Info *info, const gchar *name, const gchar *value);
 void info_add_computed_group_wo_extra(struct Info *info, const gchar *name, const gchar *value);
--- a/modules/computer.c
+++ b/modules/computer.c
@@ -1009,7 +1009,7 @@ gchar *hi_module_get_summary(void)
                     "Method=computer::getDisplaySummary\n"
                     "[%s]\n"
                     "Icon=hdd.png\n"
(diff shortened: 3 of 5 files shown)


bp0:
None of my NVMe SSDs have the exact problem of name in model, but I have some thoughts. I think the vendor match from model should happen in udisks2_util.c instead of storage.c. 

The existing code was trying to get vendor out of model to handle the problem of SATA always having ATA for vendor anyway, but for NVME, if we have a match from model we have potentially two vendors, the PCB vendor and the controller vendor. Consider "Force MP500" could match Corsair for me, and still have Phison for the controller vendor, while for your Samsung, you could have the two vendors as well, even though they're the same, as Samsung makes its own controllers. 

There is still a problem of potential double-vendors when composing the list, but that problem already existed, like with your Toshiba drives there.

bp0:
I think my solution to the double vendor problem was just to use the tags for the vendor part, and let it be double, but with the colored tags it didn't really look so bad.

Screenshot_2020-01-01_07-23-34

bp0:
Btw, using a `vendor_list` with `vendor_list_remove_duplicates_deep()` may be useful here. Vendors are keyed on their match strings so one Vendor *v->name="Samsung" may not match another Vendor *v->name="Samsung".

bp0:
Ok, sorry to spam, but I've been looking into it some more and there is already a `vendor_list` in `struct udiskd` that could be used to make a ribbon of tags with `vendor_list_ribbon()`. 

Also, I think maybe we could add a member `pcid *pci_controller` to `struct udiskd` with all the PCI information and then using code from pci.c, we get all controller vendor/subvendor and even the PCI express information along with it.

bp0:
Alright I've added the PCIe info for NVMe devices in #511. Note that it doesn't fix your double Samsung problem.

ocerman:
Yes, color vendor tag + model string seems like best solution. In the end it could look like this (please ignore that "samsung" tag is on toshiba/st drives)

new-storage

I will try to finish it tomorrow as right now it has some issues (because same list is also used in computer->hi_module_get_summary)

bp0:
Yes, I think that is the right way to go. Maybe rebase your changes off of #511 to avoid conflict. If you want I can cherry-pick your commit(s) into that PR to avoid conflicts.

ocerman:
@bp0 I have updated the code. Now it should work all fine, also fallback functions now follow same layout.

To get storage devices displayed correctly on homepage I had to create function `getStorageDevicesSimple`, but if you know better way how to do it feel free to change it. 

@bp0 Can you try to cherrypick it? I will close this PR.

bp0:
Done, this commit is now in #511.

PR #518, opened 2020-01-06
[Needs testing] New flag/funcs for escaping label part in shell keys

### LINKED ISSUE #509: Pango markup can't be used in the key if it contains an equals sign
There should be a hardinfo-standard way that keys can be escaped and then un-escaped before use in GUI. The Glib conf file parser doesn't offer any for the key side.

This prevents using vendor tags on the key side because they contain `style="..."`.

bp0:
Adds a new flag `@` to mark the label portion of the key as escaped with `key_label_escape()`. 
`key_get_components()` will automatically un-escape the label part when used.

`key_label_escape()` is just an alias for `gg_strescape()`, a modified version of glib's `g_strescape()` that allows specifying extra characters to escape, in our case "$#=" which have special meaning in the conf file format or in hardinfo. `g_strcompress()` is used as normal to un-escape it. Of course, the escape could be done manually: `=` is `\075`, for example.

Fixes #509.

Still needs some testing, there are a couple TODO:s added, so the LGTM bot will need to comment on every little change....

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,6 +281,7 @@ add_executable(hardinfo
 	hardinfo/socket.c
 	hardinfo/util.c
 	hardinfo/gg_key_file_parse_string_as_value.c
+	hardinfo/gg_strescape.c
 	hardinfo/problem_marker.c
 	hardinfo/hinote_util.c
 	hardinfo/vendor.c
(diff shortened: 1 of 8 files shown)


lpereira:
Awesome! Seems to work just fine here.

bp0:
There are still some problems regarding the disambiguator `#`, for example try flatten() -> unflatten() -> flatten() and they will be lost. When I got distracted from this I think I was planning to always escape in flatten(), but I'll have to look at it more again.

PR #526, opened 2020-01-21
util_edid: make buildable on glib<2.52

### LINKED ISSUE #525: Cannot be built on current stable version of Slax (v 9.11.0)
Seems like `g_utf8_make_valid ()` need glib >= 2.52
Slax 9 is based on Debian 9 and it comes with version 2.50

bp0:
Not a perfect fix, but moves forward.
Need to look at this more later.
Fixes #525

### DIFF
--- a/deps/sysobj_early/include/util_edid.h
+++ b/deps/sysobj_early/include/util_edid.h
@@ -25,6 +25,11 @@
 #include <stdint.h>  /* for *int*_t types */
 #include <glib.h>
 
+// TODO: find a better fix, I've seen a few EDID strings with bogus chars
+#if !GLIB_CHECK_VERSION(2,52,0)
+#define g_utf8_make_valid(S,L) g_strdup(S)
+#endif
+
 typedef struct _edid edid;
 
 typedef struct {
--- a/deps/sysobj_early/include/util_edid.h
+++ b/deps/sysobj_early/include/util_edid.h
@@ -25,11 +25,6 @@
 #include <stdint.h>  /* for *int*_t types */
 #include <glib.h>
 
-// TODO: find a better fix, I've seen a few EDID strings with bogus chars
-#if !GLIB_CHECK_VERSION(2,52,0)
-#define g_utf8_make_valid(S,L) g_strdup(S)
-#endif
-
 typedef struct _edid edid;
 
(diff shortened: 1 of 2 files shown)


deps/sysobj_early/include/util_edid.h
```diff
@@ -25,6 +25,11 @@
 #include <stdint.h>  /* for *int*_t types */
 #include <glib.h>
 
+// TODO: find a better fix, I've seen a few EDID strings with bogus chars
+#if !GLIB_CHECK_VERSION(2,52,0)
+#define g_utf8_make_valid(S,L) g_strdup(S)
```
lpereira: Might be better to declare it as a weak function, like so:

```C
gchar *g_utf8_make_valid(const gchar *s, const gsize l) __attribute__ ((weak, alias ("g_strndup")));
```

This way, if HardInfo is linked against an older glibc, but then the system is updated to a new glibc, symbol interposition will make the newer version available automatically without requiring HardInfo to be rebuilt.  (Also, it's a better idea to use `g_strndup()` here, because `g_utf8_make_valid()` takes a length argument too and that may be important if you're slicing strings.)

bp0: Yeah, that's better, but g_strndup()'s length is unsigned, but g_utf8_make_valid()'s can be negative to indicate a null-terminated string. All uses in this case are safe, however.

bp0: After trying it, it doesn't seem like the linker can overcome the type difference.

lpereira: Ah, then it might be better to implement `g_utf8_make_valid()` as a weak function (not aliasing `g_strndup()`), then:

```C
#if !GLIB_CHECK_VERSION(2, 52, 0)
__attribute__((weak))
gchar *g_utf8_make_valid(const gchar *s, const gssize l)
{
    return l < 0 ? g_strdup(s) : g_strndup(s, (gsize)l);
}
#endif
```

(Just to avoid surprises down the road.)

bp0: Alright, check out the new commit.

lpereira:
LGTM, thanks!

PR #533, opened 2020-02-03
Sensors: Groups, Icons, updated labels

ocerman:
sensors


lpereira:
Looks nice!  Where did you get those icons?  Be sure to credit the artists in the about box, too.

Sensors used to be grouped by type (rather than source of information) before, but since you have to select a sensor to paint the chart, I changed it so it's a flat list; this makes it easier for the shell to select the first item that actually matters.

Also, because there are icons in each second-level element of that tree, there's a gap before the label for the sensor driver name.  If you group by type, though, and add the sensor driver name where the type column is today, the list will look less weird.

ocerman:
> Looks nice! Where did you get those icons? Be sure to credit the artists in the about box, too.

The fan icon is from artist epicbard, I have included the credit in commit de7e460873375ec407149687dc7b14b5d89da446 The icon is under CC BY-SA 3.0 license.

The bolt icon I have created myself by cutting bolt from gnome's "battery-good-charging" icon. The gnome project is already credited in about box.

> Sensors used to be grouped by type (rather than source of information) before, but since you have to select a sensor to paint the chart, I changed it so it's a flat list; this makes it easier for the shell to select the first item that actually matters.

Well, I thought it would be best to have the list grouped by device, like CPU, GPU, MB, ... But having it grouped by source of information is close-enough. Having the list grouped helps when there is a lot of sensors.

> Also, because there are icons in each second-level element of that tree, there's a gap before the label for the sensor driver name.

Yeah, I originally wanted to have icons also for the groups. 

> If you group by type, though, and add the sensor driver name where the type column is today, the list will look less weird.

I will try that. I think in some cases having sensors grouped by device/source of information can be better though. Probably the best would be to give the user an option to choose between both groupings

bp0:
Any way you could get svg instead of png?

ocerman:
@bp0 I do not have svg of fan icon. The bolt one should be this one:
[bolt.svg.zip](https://github.com/lpereira/hardinfo/files/4174668/bolt.svg.zip)

ocerman:
I have pushed another commit that groups sensors by type. It can be changed back to group by device source by changing SENSORS_GROUP_BY_TYPE to 0.

It would be nice to later create an option in GUI so the end user could decide which option is better for him.

lpereira:
Merged, thanks!

PR #538, opened 2020-03-28
Don't count ROM as RAM in Memory Devices/Arrays

### LINKED ISSUE #536: Memory devices should not count ROM device as RAM
Screenshot_2020-03-26_21-49-34

Screenshot_2020-03-26_21-49-42

dmidecode snip for extra device:
```
Handle 0x004E, DMI type 17, 34 bytes
Memory Device
	Array Handle: 0x0026
	Error Information Handle: Not Provided
	Total Width: 16 bits
	Data Width: 16 bits
	Size: 16 MB
	Form Factor: <OUT OF SPEC>
	Set: None
	Locator: SPI ROM
	Bank Locator: Not Specified
	Type: Flash
	Type Detail: None
	Speed: Unknown
	Manufacturer: Winbond
	Serial Number: Not Specified
	Asset Tag: Not Specified
	Part Number: 25X/Q Series
	Rank: Unknown
	Configured Memory Speed: Unknown
```

bp0:
Fixes #536. Although, this is the only example I have of a dmidecode memory device that is not RAM, so I don't know how universal it is.
Screenshot_2020-03-28_00-37-00

### DIFF
--- a/deps/sysobj_early/data/vendor.ids
+++ b/deps/sysobj_early/data/vendor.ids
@@ -1278,6 +1278,13 @@ name Westinghouse Electronics
     ansi_color 0;97;44
     match_string Westinghouse
 
+name Winbond Electronics Corporation
+    name_short Winbond
+    url www.winbond.com
+    wikipedia Winbond
+    ansi_color 0;31;107
+    match_string Winbond
+
 name X.Org Foundation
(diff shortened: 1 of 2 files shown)


lpereira:
Thanks!

PR #544, opened 2020-05-07
Replace app icons with symbolic icon

runlevel5:
Peek 2020-05-07 20-16

Let's refresh the GUI with a touch of symbolic icons. These are from the GNOME Icon theme.

These icons look modern yet great for people with impaired vision.


lpereira:
Thanks for the PR, but I'm not changing the icons, especially for monochrome symbolic icons.

There are issues in using them (e.g. if you're using a dark theme they won't be that visible), and if we were to support them, I'd rather re-design the UI in such a way that the wouldn't look out of place as they look now with the rest of the UI looking like that.

runlevel5:
> There are issues in using them (e.g. if you're using a dark theme they won't be that visible), and if we were to support them, I'd rather re-design the UI in such a way that the wouldn't look out of place as they look now with the rest of the UI looking like that.

Light grey symbolic icons can be used for dark theme. gnome-control-center is one example:

Screenshot from 2020-05-08 10-48-24

However, IMHO I think there are scenarios where icon metaphors aren't expressive enough and it is best to go without.

lpereira:
Yes, light grey icons can be used with the dark theme; but they're not in this patch.  What I'm saying is: I'm not closed to the possibility of refreshing the UI, but it must be done with care.  It's not just changing icons, either; it would need a complete facelift for these new-style icons to not feel out of place.  For the next major version of HardInfo (not 0.6, which is what the current HEAD is going to be), I'm planning to junk the support for GTK+2 and focus on making it GTK3/GTK4 only; then, with the newer widgets and things like that, using symbolic icons will be indeed something very welcome.

Now, if there was a change to load icons for the current theme -- and fallback to the current icons if none were found/selected -- then I wouldn't mind integrating that patch.

runlevel5:
> I'm planning to junk the support for GTK+2 and focus on making it GTK3/GTK4 only; then, with the newer widgets and things like that, using symbolic icons will be indeed something very welcome.

:+1: I love to see GTK3/4 support. Let's revisit this topic then.

PR #547, opened 2020-05-11
Fix #545 info_add_group may goes wrong

### LINKED ISSUE #545: info_add_group in loop goes wrong
```
gchar *callback_memory_slot(void)
{
    struct Info *info = info_new();
    gchar group_title[1024];
    MemorySlot slot;
    gint i;

    if (computer->mem_slot)
    {
      for (i = 0; i < computer->mem_slot->len; ++i)
      {
          slot = g_array_index(computer->mem_slot, MemorySlot, i);
          sprintf(group_title, "%s %d", _("Memory Slot"), i + 1);

          info_add_group(info, group_title,
              info_field(_("Manufacturer"), slot.manufacturer),
              info_field(_("Locator"), slot.locator),
              info_field(_("Type"), slot.type),
              info_field(_("Spped"), slot.spped),
              info_field(_("Configured Memory Speed"), slot.configured_speed),
              info_field(_("Size"), slot.size),
              info_field_last());
      }
    }
    else
    {
        info_add_group(info, _("Error"),
            info_field(_("Error"), _("Unable to detect memory slot information")),
            info_field_last());
    }

    return info_flatten(info);
}
```

I add some code for me to show memory slots info in a single page, `info_add_group` in a for loop seems does not work well. It should display like below.

```
> memory slot 1
  ka1: va1
  ka2: va2
  ka3: va3
> memory slot 2
  kb1: vb1
  kb2: vb2
  kb3: vb3
```

but I get.

```
  ka1: va1
  ka2: va2
  ka3: va3
  ka1: va1
  ka2: va2
  ka3: va3
```

Arondight:
```
void info_add_group(struct Info *info, const gchar *group_name, ...)
{
    struct InfoGroup group = {
        .name = group_name,
        .fields = g_array_new(FALSE, FALSE, sizeof(struct InfoField))
    };
    /* ... */
}
```

Before we directly assign the second parameter to the `group.name`, but it is may gose wrong if the second parameter is a array or pointer. If we pass a pointer to the second parameter, continuously modify the value of this pointer in the loop to create multiple groups, all group will have same `group.name` ! So we assign the duplication of second parameter to the `group.name` now.

More info see #545 , google translation, may you can understand :)

### DIFF
--- a/hardinfo/info.c
+++ b/hardinfo/info.c
@@ -79,7 +79,7 @@ void info_group_add_fields(struct InfoGroup *group, ...)
 struct InfoGroup *info_add_group(struct Info *info, const gchar *group_name, ...)
 {
     struct InfoGroup group = {
-        .name = group_name,
+        .name = g_strdup(group_name),
         .fields = g_array_new(FALSE, FALSE, sizeof(struct InfoField))
     };
     va_list ap;


lpereira:
Thanks for the PR, but this doesn't seem to be a fix.

The memory will now leak everywhere that these structs are used because nothing is freeing it.  That field is supposed to point to memory that's either not freed (e.g. string literals), or memory that is managed by the info_group user.

Maybe groups should have a `free_name_on_flatten` flag like `InfoField`?

Arondight:
Yes you are right we should free these memories, and we do not need any flag if all `group.name` is created by `g_strdup`. But we can't automatically free these GArray(s) through g_array_free if we do that, things gose complex.

I've accpet this patch on my own computer, few memory leak is ok to me. I hope we can do this better (fox example rewrite `g_array_free`) ...

Hope you can understand me.

PR #557, opened 2020-06-13
Prevent compiler warnings

SamsonovAnton:
Eliminated issues:
* Implicit function declaration.
* Implicit type conversion when passing arguments to functions.
* Assignment in conditional expression.

### DIFF
--- a/includes/boots.h
+++ b/includes/boots.h
@@ -1,0 +1,6 @@
+#ifndef __BOOTS_H__
+#define __BOOTS_H__
+
+void scan_boots_real(void);
+
+#endif /* __BOOTS_H__ */
--- a/includes/computer.h
+++ b/includes/computer.h
@@ -178,4 +178,22 @@ extern GHashTable *memlabels;
 void init_memory_labels(void);
 void scan_memory_do(void);
 
+/* Scan callbacks */
+void scan_summary(gboolean reload);
+void scan_os(gboolean reload);
+void scan_security(gboolean reload);
+void scan_modules(gboolean reload);
+void scan_boots(gboolean reload);
+void scan_locales(gboolean reload);
+void scan_fs(gboolean reload);
+void scan_memory_usage(gboolean reload);
+void scan_display(gboolean reload);
+void scan_network(gboolean reload);
+void scan_users(gboolean reload);
+void scan_groups(gboolean reload);
+void scan_env_var(gboolean reload);
+#if GLIB_CHECK_VERSION(2,14,0)
+void scan_dev(gboolean reload);
+#endif /* GLIB_CHECK_VERSION(2,14,0) */
+
 #endif				/* __COMPUTER_H__ */
--- a/includes/dmi_util.h
+++ b/includes/dmi_util.h
@@ -66,4 +66,6 @@ char *dmidecode_match(const char *name, const dmi_type *type, const dmi_handle *
 
 void dmidecode_cache_free();
 
+void __scan_dmi(void);
+
 #endif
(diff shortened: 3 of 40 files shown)


includes/computer.h
```diff
@@ -178,4 +178,22 @@ extern GHashTable *memlabels;
...
+void scan_fs(gboolean reload);
+void scan_memory_usage(gboolean reload);
+void scan_display(gboolean reload);
+void scan_network(gboolean reload);
+void scan_users(gboolean reload);
+void scan_groups(gboolean reload);
+void scan_env_var(gboolean reload);
+#if GLIB_CHECK_VERSION(2,14,0)
+void scan_dev(gboolean reload);
+#endif /* GLIB_CHECK_VERSION(2,14,0) */
```
lpereira, line 197: These don't need to go in the header; these functions could be marked as `static` in `computer.c`.

includes/boots.h
```diff
@@ -0,0 +1,6 @@
+#ifndef __BOOTS_H__
+#define __BOOTS_H__
+
+void scan_boots_real(void);
```
lpereira, line 4: This can be declared in `computer.c` as well, and marked as `static`

includes/dmi_util.h
```diff
@@ -66,4 +66,6 @@ char *dmidecode_match(const char *name, const dmi_type *type, const dmi_handle *
 
 void dmidecode_cache_free();
 
+void __scan_dmi(void);
```
lpereira, line 69: This can be declared as `static` in `devices.c`. This shouldn't be in the header.

includes/dt_util.h
```diff
@@ -98,4 +98,6 @@ typedef struct {
 /* free result with g_free() */
 dt_opp_range *dtr_get_opp_range(dtr *, const char *name);
 
+void __scan_dtree(void);
```
lpereira, line 101: Likewise.

includes/gpu_util.h
```diff
@@ -58,4 +58,6 @@ void gpud_list_free(gpud *);
 
 void gpud_free(gpud *);
 
+void scan_gpu_do(void);
```
lpereira, line 61: Likewise.

includes/groups.h
```diff
@@ -0,0 +1,6 @@
+#ifndef __GROUPS_H__
+#define __GROUPS_H__
+
+void scan_groups_do(void);
```
lpereira, line 4: Likewise.

includes/inputdevices.h
```diff
@@ -0,0 +1,6 @@
+#ifndef __INPUTDEVICES_H__
+#define __INPUTDEVICES_H__
+
+void __scan_input_devices(void);
```
lpereira, line 4: Likewise.

modules/computer.c
```diff
@@ -58,24 +62,6 @@ gchar *callback_env_var(void);
...
-void scan_fs(gboolean reload);
-void scan_memory_usage(gboolean reload);
-void scan_display(gboolean reload);
-void scan_network(gboolean reload);
-void scan_users(gboolean reload);
-void scan_groups(gboolean reload);
-void scan_env_var(gboolean reload);
-#if GLIB_CHECK_VERSION(2,14,0)
-void scan_dev(gboolean reload);
-#endif /* GLIB_CHECK_VERSION(2,14,0) */
```
lpereira, line 77: Why these have been moved to the header file?  They're used only in this `.c`; it would be better if they were marked as `static` and the prototypes were kept here.

SamsonovAnton: > They're used only in this `.c`.

That is not entirely true, as `scan_os()` is called by [modules/computer/boots.c:31](https://github.com/lpereira/hardinfo/blob/3ff7143f29e332bee854c5f009cf3be550a29afc/modules/computer/boots.c#L31). The rest may be an overkill, indeed, — I just assumed those similarly looking functions might also be referenced elsewhere.

PR #558, opened 2020-06-13
E2K: added initial support of MCST Elbrus 2000 CPU architecture

EntityFX:
E2K (Elbrus 2000) - this is VLIW/EPIC architecture, like Intel Itanium (IA-64) architecture.

Ref: https://en.wikipedia.org/wiki/Elbrus_2000
image


includes/e2k/processor-platform.h
```diff
@@ -0,0 +1,74 @@
+/*
+ *    HardInfo - Displays System Information
+ *    Copyright (C) 2003-2006 Leandro A. F. Pereira <leandro@hardinfo.org>
```
lpereira, line 3: I didn't write this.  You can put your name here instead.

modules/devices/e2k/processor.c
```diff
@@ -0,0 +1,387 @@
+/*
+ *    HardInfo - Displays System Information
+ *    Copyright (C) 2003-2006 Leandro A. F. Pereira <leandro@hardinfo.org>
```
lpereira, line 3: Same thing.

modules/devices/e2k/processor.c
```diff
@@ -0,0 +1,387 @@
...
+
+#define PROC_SCAN_READ_BUFFER_SIZE 1024
+GSList *processor_scan(void)
+{
+    GSList *procs = NULL, *l = NULL;
+    Processor *processor = NULL;
+    FILE *cpuinfo;
+    gchar *buffer;
+
+    buffer = g_malloc(PROC_SCAN_READ_BUFFER_SIZE);
```
lpereira, line 250: Buffer is never freed; might be fine to just use a `char buffer[1024]`.

modules/devices/e2k/processor.c
```diff
@@ -0,0 +1,387 @@
...
+    cpuinfo = fopen(PROC_CPUINFO, "r");
+    if (!cpuinfo)
+        return NULL;
+
+    while (fgets(buffer, PROC_SCAN_READ_BUFFER_SIZE, cpuinfo)) {
+        int rlen = strlen(buffer);
```
lpereira, line 256: `fgets()` already returns the number of bytes read in that line, so you don't need to call `strlen()` here.

includes/e2k/processor-platform.h
EntityFX: Fixed

modules/devices/e2k/processor.c
EntityFX: Fixed

modules/devices/e2k/processor.c
EntityFX: Defined as `char buffer[1024]`

modules/devices/e2k/processor.c
EntityFX: Defined as char buffer[1024]

lpereira:
Merged, thanks!

EntityFX:
https://translate.google.com/translate?hl=ru&sl=ru&tl=en&u=http%3A%2F%2Fftp.altlinux.org%2Fpub%2Fpeople%2Fmike%2Felbrus%2Fdocs%2Felbrus_prog%2Fhtml%2F

PR #563, opened 2020-06-23
update README due newer libsoup requirements

mckaygerhard:


mckaygerhard:
dont merge! it need more work.. i'll check later..  seem there's some missing problems with json

PR #564, opened 2020-06-23
README updates and fix json link error due benchmarck depens links

mckaygerhard:
i dont know but seems are need.. in gtk3 does work perfect, event if soup are disabled! it need for other compilations, so i'll update README

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,9 +89,7 @@ else()
 endif()
 
 pkg_check_modules(LIBSOUP libsoup-2.4>=2.42)
-if (LIBSOUP_FOUND)
-	pkg_check_modules(JSON_GLIB REQUIRED json-glib-1.0)
-endif()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -89,9 +89,7 @@ else()
 endif()
 
 pkg_check_modules(LIBSOUP libsoup-2.4>=2.42)
-if (LIBSOUP_FOUND)
-	pkg_check_modules(JSON_GLIB REQUIRED json-glib-1.0)
-endif()
+pkg_check_modules(JSON_GLIB REQUIRED json-glib-1.0)
```
lpereira, line 92: This will only be required after #569 is merged.  I'd prefer if this change were part of that PR.

mckaygerhard: ok i'll cancel this PR and send new one later

PR #575, opened 2020-08-07
Fix man page directory

h3xx:
Use the standard variable from GNUInstallDirs.

*(Note: I don't feel a particular need to add myself to the AUTHORS file, a commit in the tree will suffice.)*


mckaygerhard:
umm cmake module depends? i means gnudirs?

lpereira:
Merged, thanks!

(About the AUTHORS file: that's entirely up to you to add your name there.  Generally, people add themselves only after they thought they contributed a significant amount of effort to the project, but what effort and significant means is up to their own interpretation.)

PR #585, opened 2020-12-12
devices/storage: Improved S.M.A.R.T. report

ocerman:
- Added interpreted S.M.A.R.T. attribute values
 - The values from the S.M.A.R.T are now printed in monotype font and aligned to the columns
 - Added more S.M.A.R.T. attribute labels (found on Samsung SSD)
 - I have also added myself to the AUTHORS file

smart


lpereira:
Nice! The SMART table was kind of bothering me before because it wasn't aligned.

PR #591, opened 2020-12-30
Fix version reading for some of development tools

kapsh:


hardinfo/util.c
```diff
@@ -1432,6 +1432,7 @@ gboolean hardinfo_spawn_command_line_sync(const gchar *command_line,
                                           GError **error)
 {
     shell_status_pulse();
+    DEBUG("Trying: %s", command_line);
```
kapsh, line 1435: Mind if I leave it here? It slightly helps to discover how hardinfo works internally.

modules/computer.c
```diff
@@ -294,7 +294,8 @@ void scan_dev(gboolean reload)
        { N_("LLDB"), "lldb --version", "(?<=lldb version )(\\d\\.?)+", TRUE },
        { N_("strace"), "strace -V", "\\d+\\.\\d+\\.?\\d*", TRUE },
        { N_("valgrind"), "valgrind --version", "\\d+\\.\\d+\\.\\S+", TRUE },
-       { N_("QMake"), "qmake --version", "\\d+\\.\\S+", TRUE},
+       { N_("QMake (Qt 5)"), "qmake-qt5 --version", "\\d+\\.\\S+", TRUE},
```
kapsh, line 297: `qmake` binary (or symlink) is not always installed, and there can be more than one present.

modules/computer.c
```diff
@@ -280,7 +280,7 @@ void scan_dev(gboolean reload)
        { N_("C (Clang)"), "clang -v", "\\d+\\.\\d+", FALSE },
        { N_("D (dmd)"), "dmd --help", "\\d+\\.\\d+", TRUE },
        { N_("Gambas3 (gbc3)"), "gbc3 --version", "\\d+\\.\\d+\\.\\d+", TRUE },
-       { N_("Java"), "javac -version", "\\d+\\.\\d+\\.\\d+", FALSE },
+       { N_("Java"), "javac -version", "(?<=javac )(\\d+.?)+", TRUE },
```
kapsh, line 283: Java is a mess. Recent versions (checked on 11) report version to stdout while older (checked 6 and 8) write it to stderr. But taking into account that 11 is only LTS available (for "casual" users) and java 8 have no public updates anymore, should be safe to use newer way from now.

Regex changed because I've seen `javac 11.0.9.1` on centos 7.

kapsh:
I noticed that some of development tools displayed as "not found" despite them being installed. This should fix detection for some of them.

2020-12-30_19-14-38_devtools

kapsh (commented):
I can't see how to fix "Python (default)" within this function: it targets stderr for `python` command but this works only for python2. When this command points to python3, it's stdout ignored.

Maybe just remove that line?

mckaygerhard:
where are qt4 related here? some 3thr party system only runs in qt4

modules/computer.c
mckaygerhard: almost 90% of 3trh party java progams does not runs with "only LTS available" seems you are talking only about openjdk right?

mckaygerhard: almost 90% of 3trh party java progams does not runs with "only LTS available", almost java 8 are commonly used today, seems you are talking only about openjdk right?

modules/computer.c
mckaygerhard: where are qt4 related here? some 3thr party system only runs in qt4

mckaygerhard:
this is a mess.. for those software that are crap in versioning (python java etc) is better to provide module for deteccion and present as comma separated string by example:

```
Java: 8 (8.3.126), 11 (11.9.1)
php: 5 (5.6.40), 7 (7.3.4)
```

lpereira:
Thanks, @kapsh!

Since you marked this PR as a draft, here are some suggestions to improve this feature:

- Instead of a boolean saying "use stderr", have a mask: "stderr" (1<<2), "stdout" (1<<0), or "stderr+stdout" (1<<2|1<<0).  If both are set, and if one produces empty output, we try the other; otherwise, we try the first one.
- Group by programming language, and list known executable names for that programming language. Useful for things like Java, Python, C and C++, and even C# (with Roslyn and Mono for instance). If multiple versions are found, we list them as @mckaygerhard suggested.

modules/computer.c
kapsh: Thanks for pointing this out. I remember qt4 being unsupported for five years, but it looks like [at least stable debian has it](https://packages.debian.org/buster/qt4-qmake), thus it should be supported here as well.

kapsh:
@lpereira @mckaygerhard I believe that your suggestions can be summarised in this way: 
1. Find every possible binary _by wildcard_. Say, `python*` => python, python2, python3 => display both binary name and version in GUI as "Python: python (version 2.7), python3 (version 3.9), etc.". Probably this will also require running glob over every `PATH` item.
2. Pack 3 values into `read_stdout` to solve stderr<=>stdout changes between releases.

Sounds fine?

modules/computer.c
kapsh: In my experience there is only single `java` binary installed inside /bin or /usr/bin in various distributions. 
Played with centos:7 container just to be sure. `/usr/bin/java` symlinked to `/etc/alternatives/java` which is symlinked to one of javas inside `/usr/lib/jvm`.

lpereira:
By wildcard might be a problem, because, let's say we glob on `python*`, and there's a program called `python-something` in the path; that'll be executed, too.

kapsh:
Yeah, thanks, that's a good point I did miss.
Wild shot: wildcards could be stricter: `python?.?, python?, python` to 
filter everything else out, maybe there is a better way too.

IOW it would be nice to avoid hardcoded values at all (as qmake family in 
current patch) while keeping code as simple as possible.

modules/computer.c
mckaygerhard: it does not matter is are supproted or not.. currently there are complete desktop build around qt4 you know it? trinity desktop!  and some NASA related programs only runs under qt4 and python2 you know it?

baryluk:
The `python` one should be left alone. It is reporting things correctly as not found.

CryptoGodfatherVA (approved):

CryptoGodfatherVA (approved):

PR #607, opened 2021-06-02
Added value check for HAS_LIBSENSORS 

wayoflife:
Tried to compile on Fedora 33 and ran into the issue that sensors.h could not be found. It turned out that HAS_LIBSENSORS was correctly defined as 0 but since it was defined, the library was expected to be present.

Changed it according to similar occurrences in the project to first check if the variable is defined and then the value.


lpereira:
Thanks for the patch. Merged!

PR #612, opened 2021-09-11
devices/dmi: Fixed DMI display when DMI is not available

ocerman:
This PR fixes following issues when the DMI is not availble:
 - the 'DMI is not available' message is now properly displayed
 - errors in stderr are no longer printed


lpereira:
Merged, thanks!

PR #618, opened 2021-10-02
devices/sensors - Added sensors from Windfarm driver

ocerman:
Added support for sensors on older Apple PowerPC computers
windfarm


lpereira:
Merged, thanks!

PR #627, opened 2021-10-28
devices/storage: Added more SMART Attributes keys

ocerman:
Added more SMART Attributes keys - so they are now translatable


lpereira:
Merged, thanks!

Issue #15, opened 2014-01-26
hardinfo displays only the first HDD's temperature in Devices>Sensors


OldTango:
Hardinfo won't display all of my Hard Drive temps. It will only show the temp for my primary drive /dev/sda but not any additional drives. I have hddtemp installed and running. I have manually added my drives to /usr/share/hddtemp/hddtemp.db. I have editd /etc/conf.d/hddtemp to include checking of all my drives.

Drives installed
"WDC WD5000HHTZ-04N21V0"  194 C "Western Digital VelociRaptor 500GB" /dev/sda
"TOSHIBA DT01ACA300"      194 C "Toshiba 3TB" /dev/sbb
"Hitachi HDS723015BLA642" 194 C "Hitachi Deskstar 1.5TB" /dev/sdc
"HDT722516DLA380"         194 C "Hitachi Deskstar 160GB" /dev/sdd

I can manually run hddtemp on any /dev/? and it reports the current temp, but hardinfo doesn't appear to pick them up or display them in Devices>Sensors

mckaygerhard:
@lpereira the output of telnet seems incorrect for spliting in code, only has split by "|" and then by "\n" ar the for.. but there's no new line char in the daemon out, so separator can be two options:
1) a double pipe "||" as can see each disk las work are "c" and then a double pipe
2) the word "/dev/" including bars

output of telnet in a large scale production server:

```
Trying 127.0.0.1...
Connected to 127.0.0.1.
Escape character is '^]'.
|/dev/sg0|MB0500EBZQA|32|C||/dev/sg1|MB0500EBZQA|32|C||/dev/sg2|TOSHIBA DT01ACA200|32|C||/dev/sg3|TOSHIBA DT01ACA200|32|C||/dev/sda|MB0500EBZQA|SLP|*||/dev/sdb|MB0500EBZQA|SLP|*||/dev/sdc|TOSHIBA DT01ACA200|32|C||/dev/sdd|TOSHIBA DT01ACA200|32|C|Connection closed by foreign host.
```

i'll try some piece of code and then paste here even made a pull request after u reviewed

kmikita:
Fix it and correct unit show.
````
---
 modules/devices/sensors.c | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/modules/devices/sensors.c b/modules/devices/sensors.c
index 696bc2d..8bcc783 100644
--- a/modules/devices/sensors.c
[...]

PR #632, opened 2021-11-18
modules/devices/sensors.c : Issue #15

kmikita:
Fixing Issue #15 https://github.com/lpereira/hardinfo/issues/15
and correct Celsius/Fahrenheit display.


lpereira:
Good catch! Will merge the patch fixing the unit later.

Issue #640, opened 2022-01-20
Benchmarking does not complete


Brishen:
Built the latest version from master on Manjaro on a Ryzen 5600X.

All benchmarks start a single thread and fail to complete.      

Upon exit, I get a number of memory leaks (see below).

Building from commit c6a5a871c2be73338e881eddf550c5433e96148f produces working benchmarks however.

```
==24807==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 61455 byte(s) in 15 object(s) allocated from:
    #0 0x7f6b51afd279 in __interceptor_malloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7f6b50cb0f81 in g_file_get_contents (/usr/lib/libglib-2.0.so.0+0x40f81)
    #2 0x7269762f73656368  (<unknown module>)

Direct leak of 29061 byte(s) in 1007 object(s) allocated from:
    #0 0x7f6b51afd279 in __interceptor_malloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7f6b50cccca9 in g_malloc (/usr/lib/libglib-2.0.so.0+0x5cca9)

Direct leak of 16896 byte(s) in 26 object(s) allocated from:
    #0 0x7f6b51afd652 in __interceptor_realloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:164
    #1 0x7f6b503f930e  (/usr/lib/libfontconfig.so.1+0x2130e)

Direct leak of 1408 byte(s) in 11 object(s) allocated from:
    #0 0x7f6b51afd652 in __interceptor_realloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:164
    #1 0x7f6b50ccd2b8 in g_realloc (/usr/lib/libglib-2.0.so.0+0x5d2b8)

Direct leak of 1042 byte(s) in 43 object(s) allocated from:
    #0 0x7f6b51afd279 in __interceptor_malloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7f6b507fcc5f in __vasprintf_internal (/usr/lib/libc.so.6+0x7dc5f)

Direct leak of 256 byte(s) in 1 object(s) allocated from:
[...]

PR #646, opened 2022-04-29
Correction for benchmarks not running on Arch

westonmyers:
L.,

I went to run hardinfo on Arch and was unable to complete any benchmarks. High CPU usage was observed and the application would stay at the benchmarking dialog for as long as you'd let it.

Decided to check if others were having issues as well and stumbled into people in the [Arch User Repository](https://aur.archlinux.org/packages/hardinfo-git) commenting along with others here in the upstream. (#640, etc)

straced it down to a module dependency issue; as seen below.

I just wanted to push in at least a quick modification for Arch peeps and open dialog here. AUR is running a patch of the below for now as the maintainer there was able to confirm the ability to run benchmarks when patched.

I'll see about testing other distros, etc. Ideally checks would be added for this module issue. There were also some aberrations seen while testing that need to be debugged as well. Starting with something is better than nothing though!

Thanks for your time,
Weston

### DIFF
--- a/modules/benchmark.c
+++ b/modules/benchmark.c
@@ -620,9 +620,7 @@ static void do_benchmark(void (*benchmark_function)(void), int entry)
         return;
 
     if (params.gui_running) {
-        gchar *argv[] = {params.argv0, "-b",           entries[entry].name,
-                         "-m",         "benchmark.so", "-a",
-                         NULL};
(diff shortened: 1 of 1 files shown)


lpereira:
Thanks! I think that module dependencies might be broken ("-a" switch to autoload dependencies) if your patch is any indication.  I'm far away from my computer but I'll take a look next week.

PR #648, opened 2022-05-04
update of vendor id

MatrixMike:
I tested the freshly built software with a Cypress USB (to PS2) adapter {keyboard and mouse}. A random test just an old device near to hand. I use Cypress PSoC devices - BLE etc so was aware of the company name change {absorption into infineon}. I have changed the vendor name for that USB id. This reflects the info at top of the Wikipedia page for Cypress. 
I enjoy using diagnostic tools and like to have them as accurate as possible. I submit this as a pull request also partly to see if this system is being monitored. 
I ran the 'bash updatepo.sh' but after this pull request. I can submit with that later.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 project(HardInfo)
-cmake_minimum_required(VERSION 2.6)
-cmake_policy(VERSION 2.6)
+cmake_minimum_required(VERSION 2.8.12)
+cmake_policy(VERSION 2.8.12)
 
 set(HARDINFO_VERSION "0.6-alpha")
 option(HARDINFO_GTK3 "Attempt to build for GTK3 (0/off for GTK2)" 0)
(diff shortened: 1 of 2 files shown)


lpereira:
Thanks for the PR!

"updatepo.sh" should only be ran if you want to update the .po files for translation after adding new strings and before a new release. Changing the vendor.ids won't require that as it's an independent mechanism; the sync server will eventually catch up automatically. I can cherry-pick myself but this might take a while as I'm a bit away from my computer this week.

MatrixMike:
Thanks for your reply. I am slightly confused as to what my actions should be now. I can wait for the robots to pick up the request and process it over the weekend.
I enjoy working on this sort of stuff.
Best regards and for whatever it is worth ‘happy mothers’ day. No happy mother near me but you might have a good one.

lpereira:
I've merged this patch, but only the changes to the vendors file (plus the change in the AUTHORS.md). I have not merged the changes to the po files as they haven't been changed by the vendor file change.

PR #651, opened 2022-07-23
Update e2k

afonot:
Redesigned display of information about e2k processors, added information about new processors


lpereira:
Merged, thanks!  (There's just one minor nit: `icons` is allocated and used but it's never set, but I'm merging it anyway.)

PR #686, opened 2023-02-16
Update loongarch architecture processor information

zhuyaliang:
Update ```loongarch``` architecture processor information
The previous ```LoongArch``` support added in commit 56e590d was based on the kernel behavior before the official upstream merge. Stable cpuinfo: https://git.kernel.org/torvalds/c/7153c3c


lpereira:
Merged, thank you!

PR #690, opened 2023-03-01
Devices/SPD: Improved SPD decoding for DDR2

ocerman:
- First commit adds manufacturing date decoding for DDR2
  - also, the date decoding logic is put to one function as same logic is also used for DDR3/DDR4 RAM
- Second commit fixes part number decoding for DDR2 (and older) RAM.
  - As first valid part number location is at b. 73, the `*bytes` needs to be incremented to this value before `while` condition is being evaulated


lpereira:
MErged, thanks!

